PR #104, opened 2026-04-13
Update: chunk Qwen3 decode scope1 projections

ndleslx:
## Summary
- Apply the wider scope1 reduction chunk and chunked Q/KV projection loops to both `examples/models/qwen3/qwen3_32b_decode_scope1.py` and the scope1 section of `examples/models/qwen3/qwen3_32b_decode.py`.
- Keep the full decode scope3 path on its original `K_CHUNK = 128` by introducing a scope1-specific chunk constant in `qwen3_32b_decode.py`.
- Scope1 benchmark on `a2a3` device `1` with runtime profiling: `origin/main` wall time `525.04 us` and `161` tasks; updated branch wall time `350.02 us` and `37` tasks.
- Full decode benchmark on `a2a3` device `1` with `--max-seq --runtime-profiling`: before change wall time `3198.22 us` and `1503` tasks; after change wall time `3080.66 us` and `1379` tasks.

## Related Issues
- #81

### DIFF
--- a/examples/models/qwen3/qwen3_32b_decode.py
+++ b/examples/models/qwen3/qwen3_32b_decode.py
@@ -42,7 +42,7 @@ EPS = 1e-6
 HIDDEN_INV = 1.0 / HIDDEN
 
 # Scope 1 tiling constants.
-K_CHUNK = 128
+SCOPE1_K_CHUNK = 512
 Q_OUT_CHUNK = 64
 KV_OUT_CHUNK = 64
 BATCH_TILE = 16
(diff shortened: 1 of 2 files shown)


examples/models/qwen3/qwen3_32b_decode.py
```diff
@@ -42,7 +42,7 @@
 HIDDEN_INV = 1.0 / HIDDEN
 
 # Scope 1 tiling constants.
-K_CHUNK = 128
+SCOPE1_K_CHUNK = 512
```
bumble0918, line 45: 这名字有点怪

PR #167, opened 2026-04-24
Qwen3-32B decode scope1: switch Q/K/V projection to pl.spmd

yanghaoran29:
Shard Q/K/V output columns with for ob in pl.range(0, ..., SPMD_CORES) and
for i in pl.spmd(4). Assert q_out_blocks and kv_out_blocks are divisible by
SPMD_CORES; assert SPMD_CORES == 4 so outer strides cannot diverge from the
pl.spmd literal.

Expand docstrings on the program builder, tensor specs, golden helper, and inner
initializers.

Known limitation: pl.spmd does not accept a variable argument yet; inner loops
use the integer literal pl.spmd(4) until the frontend allows parameterizing the
SPMD width.


yanghaoran29:
@coderabbitai review

PR #197, opened 2026-04-28
Add DeepSeek-V4 o_proj fused kernel example

### LINKED ISSUE #196: Add DeepSeek-V4 o_proj kernel example to pypto-lib
## Summary

Port the DeepSeek-V4-Pro decode-time output projection (`o_proj`) to
`pypto-lib` as a runnable example kernel under
`examples/models/deepseek/v4/`.

`o_proj` is a LoRA-style **two-stage grouped projection** that replaces a
single dense `[H*head_dim, D]` matmul with a low-rank pair of grouped
projections (`wo_a`, `wo_b`). It is the per-layer attention output path
used during decode, and is a non-trivial pattern that differs from the
existing DeepSeek-V3 / V3.2 examples (which use either a plain
`out_proj` or an MLA-specific path).

## Reference

HuggingFace upstream:
<https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro/blob/main/inference/model.py>

`Attention.forward`, lines 537-541:

```python
o = o.view(bsz, seqlen, n_local_groups, -1)
wo_a = self.wo_a.weight.view(n_local_groups, o_lora_rank, -1)
o = torch.einsum("bsgd,grd->bsgr", o, wo_a)
x = self.wo_b(o.flatten(2))
```

This decomposes into:

- **Stage A** (per-group LoRA-down):
  `o_r = einsum("tgd,grd->tgr", o.view(T, G, D_in), wo_a)`
- **Stage B** (cross-group up-projection):
  `out = o_r.flatten(1) @ wo_b.T`

with default decode shapes (B=16, S=1, H=128, head_dim=512):
`T=16, G=16, D_in=4096, R=O_LORA=1024, D=7168`.

## Goal

Add a single fused PyPTO kernel + golden + ST entry that:

- Takes `(o, wo_a, wo_b)` as program inputs and writes `attn_out`
  directly, with the Stage-A intermediate `o_r` allocated as an internal
  on-chip workspace (mirrors the local Python tensor on model.py L540).
- Uses FP32 accumulation throughout both matmul stages
  (PTO `TileType::Acc` only supports `float` / `int32_t`; see
  `pto-isa/tmatmul.md`).
- Ships a torch-einsum golden that matches the upstream einsum
  expression exactly (no algebraic re-derivation), and a pytest ST
  entry under `tests/st/examples/deepseek_v4/`.
- Runs and matches the golden on a2a3 NPU at `rtol=atol=3e-3` for the
  default decode shapes above.

## Non-goals

- Fitting `o_proj` into the existing fused-attention examples (kept as a
  standalone kernel for now, in line with the other `deepseek_v4_*`
  decode drafts).
- Any prefill / variable-seqlen path. This issue covers decode shapes only.
- Building a generic LoRA / grouped-einsum library primitive in
  `pypto-lib`. We do this kernel as a model example first; lifting it to
  a reusable primitive is out of scope.

## Implementation notes

- Layout on the program boundary:
  - `o`     : `[G*T, D_in]`  group-major rows (row `g*T+t` =
              attention output `o_attn[t, g*D_in:(g+1)*D_in]`).
  - `wo_a`  : `[G*D_in, R]`  K-major.
  - `wo_b`  : `[D, G*R]`     RowParallelLinear weight, consumed via
              `b_trans=True` in Stage B.
  - `attn_out` : `[T, D]`.
- This packing matches the v3.2 `grouped_gemm_bf16` convention and
  guarantees every K-dimension `pl.slice` of `o` starts at column 0,
  which is necessary on a2a3 because of an unrelated backend codegen
  bug with column-strided slices of wide tensors
  (#1212 — found while bringing this kernel up; layout is the
  workaround, not the feature).
- Tile shapes (decode):
  - Stage A: `M=T=16`, `K_CHUNK=128`, `N_CHUNK=128` over `D_in=4096`.
  - Stage B: `M=T=16`, `K_CHUNK=128`, `N_CHUNK=256` over `D=7168`,
    `b_trans=True`.

## Acceptance criteria

- [ ] `examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py` builds
      and runs standalone.
- [ ] `tests/st/examples/deepseek_v4/test_o_proj_sim.py::test_v4_o_proj_fused_sim`
      passes on a2a3 NPU with `rtol=atol=3e-3`.
- [ ] Golden is the direct torch-einsum form of model.py L537-541
      (no manual re-derivation).

## Related

- Backend bug found while bringing this kernel up: #1212
  (a2a3 column-strided matmul slice miscompile). Worked around by the
  group-major-row layout described above; not a blocker for this
  feature.

zhaozhaozz:
## Summary
- Add a runnable PyPTO example for the DeepSeek-V4 decode-time `o_proj` fused kernel.
- Remove the legacy `deepseek_v4_decode_o_proj_draft.py` now that the canonical example lives at `deepseek_v4_decode_o_proj.py`.
- Retarget the remaining DeepSeek-V4 draft orchestrators to import the canonical `golden_deepseek_v4_decode_o_proj` helper.

## Kernel structure
- Prologue: pack `o[T, H, HEAD_DIM]` into `o_packed[G*T, O_GROUP_IN]`.
- Stage A: per-group `o_packed[T, K] @ wo_a[1, R, K]^T -> [1, T, R]`, then reshape/store into the grouped BF16 workspace `o_r[T, G*R]`.
- Stage B: `o_r[T, G*R] @ wo_b[D, G*R]^T -> attn_out[T, D]`.

## Notes
- `wo_a` stays in the upstream `[G, R, O_GROUP_IN]` layout; this version no longer needs the earlier internal 2D `wo_a_2d` workaround.
- Stage A and Stage B both peel the first K chunk outside the loop so only the accumulator is loop-carried.
- The golden follows the official DeepSeek-V4 `Attention.forward` o_proj dataflow: `o.view(bsz, seqlen, groups, -1)`, `torch.einsum("bsgd,grd->bsgr", o, wo_a)`, then `wo_b(o.flatten(2))`, with the intermediate matching the BF16 workspace used by the kernel.
- The checked-in example uses downscaled demo defaults and keeps the v4-pro sizes in comments.

## Validation
- Local a2a3sim:
  - `PYTHONPATH=/home/bot/code/PTO/pypto-lib python examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py --platform a2a3sim`
- Precision snapshot (seed=0, `attn_out`, shape `16x4096`):
  - `max_abs = 0.00048828`
  - `mean_abs = 0.00000012`
  - `rmse = 0.00000514`
  - `mismatches @ rtol=atol=1e-3 = 0 / 65536`
- Workflow-level local gate:
  - `/home/bot/code/PTO/local-github-ci/local-ci --repo-root "$(pwd)" --job pre-commit`

Closes #196


examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
```diff
@@ -0,0 +1,207 @@
...
+                    with pl.at(level=pl.Level.CORE_GROUP):
+                        xa0 = pl.slice(o,    [T, A_K_CHUNK],          [row_base_o, 0])
+                        wa0 = pl.slice(wo_a, [A_K_CHUNK, A_N_CHUNK],  [w_row_g, n0])
+                        acc_a = pl.matmul(xa0, wa0, out_dtype=pl.FP32)
+
+                        for kb in pl.range(1, A_K_BLOCKS):
+                            k0 = kb * A_K_CHUNK
+                            xa_chunk = pl.slice(o,    [T, A_K_CHUNK],         [row_base_o, k0])
+                            wa_chunk = pl.slice(wo_a, [A_K_CHUNK, A_N_CHUNK], [w_row_g + k0, n0])
+                            acc_a = pl.matmul_acc(acc_a, xa_chunk, wa_chunk)
```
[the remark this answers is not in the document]
zhaozhaozz: Thanks — agreed the slice on `o` is invariant within the inner `nb` loop. For this example we're keeping it explicit in-loop because (a) `o` is small (`[T, O_GROUP_IN] = [16, 4096]` BF16 = 128 KiB) so re-reading is cheap relative to the `wo_a` slice (8× larger per chunk), and (b) the two-block `pl.at(CORE_GROUP)` structure (see threads on lines 129/152) is what triggers the implicit format reformat we rely on; hoisting the slice would require restructuring those scopes. Since this file is a model example rather than a perf-tuned kernel, leaving it explicit. Will revisit if/when we promote it to a tuned op.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -0,0 +1,207 @@
...
+                            xa_chunk = pl.slice(o,    [T, A_K_CHUNK],         [row_base_o, k0])
+                            wa_chunk = pl.slice(wo_a, [A_K_CHUNK, A_N_CHUNK], [w_row_g + k0, n0])
+                            acc_a = pl.matmul_acc(acc_a, xa_chunk, wa_chunk)
+
+                    with pl.at(level=pl.Level.CORE_GROUP):
+                        o_r = pl.assemble(
+                            o_r,
+                            pl.cast(acc_a, target_type=pl.BF16),
+                            [0, out_col_g + n0],
+                        )
```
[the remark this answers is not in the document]
zhaozhaozz: Tried this — looks like pure refactor but it actually breaks codegen on a2a3. Verified empirically: merging the two blocks fails at runtime with `TStore.hpp:507 Src and dst layout must be same, only support ND/DN/NZ`.

Reason: the cube `matmul`/`matmul_acc` output sits in a ColMajor cube buffer, and `pl.cast(..., BF16)` keeps the layout. With the `assemble` in a separate `pl.at(CORE_GROUP)` scope, the framework inserts an implicit reformatting move so the tile is RowMajor before the GM (ND) store. Once the two scopes are fused, that reformat is dropped and the ColMajor → ND store mismatches.

Keeping the two blocks separate (and added an inline comment explaining this) is required for correctness, not just style. Marking this as wontfix.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
```diff
@@ -0,0 +1,207 @@
...
+                        xb_chunk = pl.slice(o_r,  [T, B_K_CHUNK],         [0, k0])
+                        wb_chunk = pl.slice(wo_b, [B_N_CHUNK, B_K_CHUNK], [n0, k0])
+                        acc_b = pl.matmul_acc(acc_b, xb_chunk, wb_chunk, b_trans=True)
+
+                with pl.at(level=pl.Level.CORE_GROUP):
+                    attn_out = pl.assemble(
+                        attn_out,
+                        pl.cast(acc_b, target_type=pl.BF16),
+                        [0, n0],
+                    )
```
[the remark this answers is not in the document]
zhaozhaozz: Same situation as the Stage A thread above (https://github.com/hw-native-sys/pypto-lib/pull/197#discussion_r3153055094): merging triggers `TStore.hpp:507 Src and dst layout must be same` on a2a3 because the implicit ColMajor → RowMajor reformat between the cube accumulator and the ND GM store is dropped. Keeping the two `pl.at(CORE_GROUP)` blocks separate (with an explanatory comment) is required for correctness.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
```diff
@@ -0,0 +1,207 @@
...
+
+    o    = tensors["o"].float()                            # [G*T, O_GROUP_IN]
+    wo_a = tensors["wo_a"].float()                         # [G*O_GROUP_IN, O_LORA]
+    wo_b = tensors["wo_b"].float()                         # [D, G*O_LORA]
+
+    # Recover [T, G, O_GROUP_IN] from group-major-row packing.
+    o_g    = o.view(O_GROUPS, T, O_GROUP_IN).transpose(0, 1).contiguous()  # [T, G, d]
+    wo_a_3 = wo_a.view(O_GROUPS, O_GROUP_IN, O_LORA)       # [G, d, R]
+    o_r    = torch.einsum("tgd,gdr->tgr", o_g, wo_a_3)     # [T, G, O_LORA]
+    out    = o_r.flatten(1) @ wo_b.T                       # [T, D]
```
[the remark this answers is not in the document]
zhaozhaozz: Done. Rewrote the golden to follow the upstream einsum contract `bsgd,grd->bsgr` (matching `inference/model.py`), and added the intermediate `o_r = o_r.to(bfloat16).float()` cast so the FP32 reference mirrors the kernel's BF16 quantization between Stage A and Stage B. Test passes on a2a3 with `rtol=atol=3e-3`.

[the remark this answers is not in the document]
zhaozhaozz: Done. Golden now reconstructs `o_g: [B,S,G,d]` and `wo_a: [G,R,d]` and uses `torch.einsum("bsgd,grd->bsgr", ...)` exactly per upstream, then casts to BF16 before the `wo_b` matmul to match the kernel's intermediate quantization. The reshape-as-2D path is gone.

zhaozhaozz:
Rebased onto latest main (adopted #194's downscaled V4 defaults: B=16, D=4096, H=64, O_GROUPS=8, with v4-pro values in comments). Addressed all review threads inline:

- gemini #1 (hoist o slice): declined — kept explicit; this is a model example, slice on small `o` is cheap, and restructuring conflicts with the CORE_GROUP scope split required by #2/#3.
- gemini #2 / #3 (merge `pl.at(CORE_GROUP)` blocks in Stage A/B): declined — verified empirically that merging breaks codegen on a2a3 (`TStore.hpp:507 Src and dst layout must be same`). The separate scopes induce an implicit ColMajor → RowMajor reformat between the cube accumulator and the ND GM store; merging drops it. Added inline comments.
- gemini #4 (BF16 intermediate in golden): accepted.
- coderabbit #5 (upstream einsum `bsgd,grd->bsgr` in golden): accepted; rewrote golden to mirror `inference/model.py`.
- coderabbit nitpick (divisibility asserts on tile counts): accepted.

Re-validated: `tests/st/examples/deepseek_v4/test_o_proj_sim.py::test_v4_o_proj_fused_sim[a2a3]` PASSED on real NPU.

zhaozhaozz:
Updated the kernel signature so it matches the data-flow contract documented in `deepseek_v4_decode_single_layer.md`:

- `o`: now `[T, H, HEAD_DIM]` BF16 — verbatim attention output, no caller-side reshape needed.
- `wo_a`: `[O_GROUPS * O_GROUP_IN, O_LORA]` BF16 — load-time pre-shuffled from upstream `[G, R, O_GROUP_IN]`. Static weight, so the reshape is a one-shot model-load cost. Tried the upstream 3D shape directly but it produced an N-strided cube operand that mis-compiled on a2a3.
- `wo_b`, `attn_out`: unchanged.

Internally, the kernel now has a small prologue that packs `o[t, g*hpg:(g+1)*hpg, :]` into a `[G*T, O_GROUP_IN]` workspace before Stage A. This is a workaround for #1212 — direct strided indexing of `o` mis-compiles on a2a3 (only row 0 comes back correct). When #1212 is fixed the prologue can collapse to an in-place view.

`single_layer.md` updated to note the `wo_a` load-time shuffle. Re-validated on a2a3: `test_v4_o_proj_fused_sim[a2a3]` PASSED.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
```diff
@@ -0,0 +1,295 @@
...
+
+            # Internal workspace bridging Stage A and Stage B. Mirrors the
+            # local Python tensor ``o_r`` produced on model.py line 540 by
+            # ``torch.einsum("bsgd,grd->bsgr", ...)``.
+            o_r = pl.create_tensor([T, GR], dtype=pl.BF16)
+
+            # ---- Prologue: pack o[t, g*hpg:(g+1)*hpg, :] -> o_packed[g*T+t, :] ----
+            # Each per-token slice [hpg, HEAD_DIM] = O_GROUP_IN BF16 is contiguous
+            # in the source, so the copy is a single DMA per (g, t).
+            for g in pl.range(O_GROUPS):
```
zhangqi-chen: change to parallel if parallel if available

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -0,0 +1,295 @@
...
+Stage-A RHS slice is ``wo_a[g, r0:r1, d0:d1]`` consumed with
+``b_trans=True``; on a2a3 this exercises the RHS side of #1212. The
+kernel therefore first views ``wo_a`` as a contiguous 2D
+``[G*R, O_GROUP_IN]`` tensor and slices ``[R_CHUNK, K_CHUNK]`` panels
+from that view. TODO(#1212): remove this reshape workaround once grouped
+strided RHS matmul is fixed.
+
+FP32 accumulator is mandatory throughout: the ISA's TileType::Acc only
+supports ``float`` or ``int32_t`` (pto-isa: tmatmul.md).
+"""
```
zhangqi-chen, line 14: delete most of the header description

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
```diff
@@ -0,0 +1,295 @@
...
+                row_base_o = g * T
+                w_row_g    = g * O_LORA
+                out_col_g  = g * O_LORA
+
+                for nb in pl.range(A_N_BLOCKS):
+                    n0 = nb * A_N_CHUNK
+
+                    with pl.at(level=pl.Level.CORE_GROUP):
+                        xa0 = pl.slice(o_packed, [T, A_K_CHUNK],         [row_base_o, 0])
```
zhangqi-chen: pl.slice -> a=b[:]

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
```diff
@@ -0,0 +1,295 @@
...
+                            wa_chunk = pl.slice(
+                                wo_a_2d, [A_N_CHUNK, A_K_CHUNK], [w_row_g + n0, k0]
+                            )
+                            acc_a = pl.matmul_acc(acc_a, xa_chunk, wa_chunk, b_trans=True)
+
+                    # Separate CORE_GROUP scope so the BF16 cast lands on the
+                    # vector path and the GM store sees a row-major tile (the
+                    # matmul accumulator is column-major in the cube buffer).
+                    with pl.at(level=pl.Level.CORE_GROUP):
+                        o_r = pl.assemble(
```
zhangqi-chen: pl.assemble -> a[:] = b

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -0,0 +1,295 @@
...
+                            o_r,
+                            pl.cast(acc_a, target_type=pl.BF16),
+                            [0, out_col_g + n0],
+                        )
+
+            # ---- Stage B: attn_out = o_r @ wo_b^T  ----
+            for nb in pl.range(B_N_BLOCKS):
+                n0 = nb * B_N_CHUNK
+
+                with pl.at(level=pl.Level.CORE_GROUP):
```
zhangqi-chen, line 122: change to this format:
acc = create_tensor
pl.at
for i in range
a=
b=
if i==0: matmul
else: matmul_acc(acc)

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
zhaozhaozz: Handled in 8098e0d. Changed the prologue group loop from `pl.range(O_GROUPS)` to `pl.parallel(0, O_GROUPS, 1)`. I intentionally did not add `chunk=...` here because parser only allows chunked `pl.parallel` inside an appropriate `pl.at(..., optimizations=[pl.auto_chunk])` context; this outer loop is not in that context.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
zhaozhaozz: Handled in 8098e0d. Shortened the module header substantially; it now only states the fused o_proj formula, the public tensor layout, and the #1212 workaround at a high level. The detailed rationale stays in the PR/notes instead of the source header.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
zhaozhaozz: Handled in 8098e0d. Replaced the explicit `pl.slice(...)` calls in the prologue, Stage A, and Stage B with tensor subscript slice syntax, e.g. `x[a:b, c:d]`.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj.py
zhaozhaozz: Handled in 8098e0d. Replaced the `pl.assemble(...)` writes with subscript-write syntax (`dst[...] = src`) for `o_packed`, `o_r`, and `attn_out`.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
zhaozhaozz: Not changed as requested after testing. The suggested `acc = create_tensor; for kb; if kb == 0: matmul else: matmul_acc` form works on a5sim, but fails for this PR target on a2a3 because PTOAS rejects the generated `tload -> loc=acc` path. A zero-initialized `pl.full(..., 0)` variant fails on both a2a3 and a5 because it generates `texpands -> loc=acc`. I kept the peeled form (`acc = matmul(first tile)` followed by `for kb in range(1, ...): matmul_acc`) because the accumulator is then produced by a legal cube matmul, it works across a2a3/a5, and it avoids the loop-internal `if` branch. Re-validated with a2a3sim.

zhaozhaozz: Updated in 7047f56. I revisited this against the Qwen3 decode pattern and changed both Stage A and Stage B to the requested `create_tensor + pl.pipeline(0, ..., stage=2) + if first matmul else matmul_acc` form. The earlier concern was too broad: the a2a3 issue was with the non-pipeline/range variant I had tested, not with this pipeline accumulator pattern. Re-validated with `python examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py -p a2a3sim`; `attn_out` passes.

zhaozhaozz:
Latest update:

- Rebased the branch onto current `origin/main` (`59668d9`).
- Kept the implementation under `deepseek_v4_decode_o_proj_draft.py` for now, per review/inspection preference.
- Addressed the accumulator-style review by switching both o_proj matmul stages to the Qwen3-style `create_tensor + pl.pipeline(0, ..., stage=2) + if first matmul else matmul_acc` pattern.
- Validation: `pre-commit run --files examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py` passed; `PYTHONPATH=$PWD PTOAS_ROOT=/home/bot/code/PTO/ptoas-bin PTO_ISA_ROOT=/home/bot/code/PTO/pto-isa python examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py -p a2a3sim` passed with `attn_out PASS`.

Note: the earlier statement that this accumulator form was problematic on a2a3 was too broad. The failing case was the non-pipeline/range variant; the `pl.pipeline` form used by Qwen3 works here as expected.

zhaozhaozz:
Follow-up update: removed the non-essential `deepseek_v4_decode_single_layer.md` changes from the PR. The PR diff now only contains the draft implementation file, so the single-layer flow document keeps the canonical `o_proj.py` name and original wording.

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -6,8 +6,16 @@
 # INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
 # See LICENSE in the root of the software repository for the full text of the License.
 # -----------------------------------------------------------------------------------------------------------
-"""DeepSeek-V4 grouped output projection (decode): grouped wo_a low-rank reduction followed by
-wo_b expansion back to model dim."""
+"""DeepSeek-V4 grouped output projection fused example.
```
zhangqi-chen, line 9: delete _draft.py

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -21,37 +29,160 @@
...
         ):
-            # TODO: kernel implementation
+            # Group-major packed view of ``o``. See module docstring for the
+            # rationale (workaround for hw-native-sys/pypto#1212).
+            o_packed = pl.create_tensor([O_GROUPS * T, O_GROUP_IN], dtype=pl.BF16)
+
+            # 2D view of upstream wo_a[g, r, d] as wo_a_2d[g*R + r, d].
+            # TODO(#1212): use the direct 3D slice once a2a3 handles grouped
+            # strided RHS matmul correctly.
+            wo_a_2d = pl.reshape(wo_a, [GR, O_GROUP_IN])
```
zhangqi-chen: use 3d slice in matmul incore

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -21,37 +29,160 @@
...
+                head_off = g * HEADS_PER_GROUP
+                row_base = g * T
+                for t in pl.range(T):
+                    with pl.at(level=pl.Level.CORE_GROUP):
+                        src_3d = o[t:t + 1, head_off:head_off + HEADS_PER_GROUP, :]
+                        src_2d = pl.reshape(src_3d, [1, O_GROUP_IN])
+                        o_packed[row_base + t:row_base + t + 1, :] = src_2d
+
+            # ---- Stage A: o_r[t, g, r] = sum_d o_packed[g*T+t, d] * wo_a[g, d, r] ----
+            for g in pl.range(O_GROUPS):
```
zhangqi-chen: change to parallel

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -21,37 +29,160 @@
...
+                row_base_o = g * T
+                w_row_g    = g * O_LORA
+                out_col_g  = g * O_LORA
+
+                for nb in pl.range(A_N_BLOCKS):
+                    n0 = nb * A_N_CHUNK
+
+                    with pl.at(level=pl.Level.CORE_GROUP):
+                        acc_a = pl.create_tensor([T, A_N_CHUNK], dtype=pl.FP32)
+                        for kb in pl.pipeline(0, A_K_BLOCKS, stage=2):
```
zhangqi-chen: use range

examples/models/deepseek/v4/deepseek_v4_decode_o_proj_draft.py
```diff
@@ -21,37 +29,160 @@
 HEAD_DIM   = 512
 O_LORA     = 1024
 O_GROUPS   = 8                  # v4-pro 16
-O_GROUP_IN = H * HEAD_DIM // O_GROUPS
+HEADS_PER_GROUP = H // O_GROUPS          # 8 heads per group
+O_GROUP_IN = HEADS_PER_GROUP * HEAD_DIM  # 4096 (matches v4-pro by coincidence)
+GR         = O_GROUPS * O_LORA           # 8192 (v4-pro 16384)
```
zhangqi-chen: delete

PR #246, opened 2026-05-11
DSv4: implement decode HCA orchestration (ratio=128 path)

yanghaoran29:
Promote deepseek_v4_decode_hca from draft to active kernel and fill in the @pl.jit orchestration body covering hc_pre + qkv_proj_rope + main compressor (ratio=128) + deterministic compress_topk_idxs + fused sparse_attn_with_proj + hc_post. Mirrors the SWA layout: split ori/cmp KV pools to match sparse_attn's static shape contract (ORI_BLOCK_NUM=B, CMP_BLOCK_NUM=B*64), pad topk_idxs to WIN+IDX_TOPK=640 with seqused_kv bounding the valid range, and wire in W8A8 wq_b/wo_b with per-channel scales plus INT8 qr/qr_scale on the (HCA-unused) qr path.


yanghaoran29:
@coderabbitai resume

PR #357, opened 2026-05-22
Add Cayley-Hamilton triangular inverse example

ChristosMatzoros:
# Add Cayley-Hamilton triangular inverse example

## Summary

Adds `examples/intermediate/tri_inverse.py` — a new intermediate example that
computes `inv(I - A)` for strict-lower-triangular `A` via the Cayley-Hamilton
doubling algorithm. Targets the chunked-GDN triangular-inverse used in
Qwen3-Next gated delta-rule attention.

## Algorithm

`A` is strict-lower-triangular, therefore nilpotent (`A^n = 0`), so the
doubling iteration terminates exactly in `ceil(log2(n))` steps:

```
X_0 = I,  Y_0 = A
for k in range(ceil(log2(n))):
    X_{k+1} = X_k + X_k @ Y_k     # equivalently X_k (I + Y_k)
    Y_{k+1} = Y_k @ Y_k
# Result: X = inv(I - A)
```

### DIFF
--- a/examples/intermediate/tri_inverse.py
+++ b/examples/intermediate/tri_inverse.py
@@ -46,12 +46,24 @@ def build_tri_inverse_program(
     m_tile: int = M_TILE,
     k_tile: int = K_TILE,
     m_chunk: int = M_CHUNK,
+    refinement_steps: int = 1,
 ):
     """Build the @pl.program for n x n triangular inverse.
 
     Specialises the doubling-step count via ceil(log2(n)).
+
+    ``refinement_steps`` runs that many Newton iterative-refinement steps
+    after the doubling loop: at each step ``X <- X + X (I - (I - A) X)``,
+    which is quadratically convergent and crushes the FP16-cube roundoff
+    error from O(eps) to O(eps^2).  Default 1 — bring n=64 inside the 2e-2
+    tolerance and tighten accuracy at all sizes.  Set to 0 for raw speed
+    when O(eps) accuracy is acceptable.
     """
     n_steps = max(1, (n - 1).bit_length())  # 7 for n=128
+    if n % m_tile != 0:
+        raise ValueError(f"n ({n}) must be a multiple of m_tile ({m_tile})")
+    if n % k_tile != 0:
+        raise ValueError(f"n ({n}) must be a multiple of k_tile ({k_tile})")
     k_blocks = n // k_tile  # 2 for n=128, k_tile=64
 
     @pl.program
@@ -79,15 +91,18 @@ def build_tri_inverse_program(
             [m_tile, n] @ [n, n] matmul, fitting comfortably in the 192 KB
             Vec budget.
 
-            Y_temp snapshots Y before Y = Y @ Y: reading the full Y while
-            another parallel iter writes to its own row slab would race;
-            snapshotting first makes the matmul read-only on Y_temp and
-            write-only on Y_state.
+            Y is double-buffered across two tensors (Y_a, Y_b) so the
+            Y = Y @ Y step reads from one buffer and writes to the other.
+            Without double-buffering the matmul would read the full Y while
+            other M-parallel iters write to row slabs of that same tensor,
+            which races; reading and writing distinct tensors removes the
+            race without needing an explicit GM-to-GM snapshot copy.
             """
             X_state = pl.create_tensor([n, n], dtype=pl.FP32)
-            Y_state = pl.create_tensor([n, n], dtype=pl.FP32)
-            Y_temp = pl.create_tensor([n, n], dtype=pl.FP32)
+            Y_a = pl.create_tensor([n, n], dtype=pl.FP32)
+            Y_b = pl.create_tensor([n, n], dtype=pl.FP32)
             X_acc = pl.create_tensor([n, n], dtype=pl.FP32)
+            R_state = pl.create_tensor([n, n], dtype=pl.FP32)  # newton refinement residual
 
             with pl.at(level=pl.Level.CORE_GROUP,
                        optimization=pl.chunked_loop_optimizer,
@@ -96,9 +111,14 @@ def build_tri_inverse_program(
                     id_row = pl.slice(identity, [m_tile, n], [mb, 0])
                     a_row = pl.slice(A, [m_tile, n], [mb, 0])
                     X_state = pl.assemble(X_state, id_row, [mb, 0])
-                    Y_state = pl.assemble(Y_state, a_row, [mb, 0])
+                    Y_a = pl.assemble(Y_a, a_row, [mb, 0])
 
             for step in pl.unroll(n_steps):
+                # Double-buffered Y: each step reads Y_a, writes Y_b, then
+                # swaps the Python-level bindings so the next iteration reads
+                # the freshly-written buffer.  No explicit snapshot/copy scope
+                # needed — the read and write target distinct GM tensors.
+                #
                 # Split X update: matmul first (writes X_acc), then add(X, X_acc).
                 # Keeping x_row + acc + x_new alive in one scope is the dominant
                 # Vec cost at n=256 (3 row tiles of size m_tile*n*4 bytes); splitting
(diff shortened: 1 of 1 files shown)


ChristosMatzoros:
## Performance analysis — engine utilization & host overhead breakdown

### What the kernel does

We compute `inv(I − A)` for strict-lower-triangular `A` (a nilpotent matrix,
so `A^n = 0`) via the Cayley–Hamilton doubling identity:

```
X_{k+1} = X_k + X_k @ Y_k        # accumulate Σ Y^k into X
Y_{k+1} = Y_k @ Y_k              # square Y each step
```

Starting from `X_0 = I`, `Y_0 = A`, after `⌈log₂(n)⌉ = 7` iterations (at
n=128) `Y → 0` and `X → inv(I − A)`. All three multiplications per iter
run on the **cube engine** via `pl.matmul` (X@Y, Y@Y) and a fused
`pl.matmul_acc` for the `+X` term: `L0C += X @ I`, where reloading `X` as
the Left operand into L0A and using `I` as the Right operand into L0B
lets us fold `+X` into the existing `X@Y` accumulator in a single L0C
slot — avoiding the unsupported **L0C→Mat tile-move** on a2/a3 (which is
also why X and Y cannot stay in L0 across iterations and must round-trip
to GM at the end of every doubling step). The memory flow per iter is
`GM → L1 Mat → L0A/L0B → cube → L0C → GM`, with `pl.range(n_steps,
init_values=(X, Y))` threading the GM tensor handles for X and Y across
iterations as iter_args. The seeds `X_buf ← I` and `Y_buf ← A` are also
laid down via AIC-only matmul detours (`I@I` and `A@I`) so the data path
stays on the cube engine — a direct `Mat → GM` store would route through
the Vec engine and spawn a 512 KB cube-to-vec pipe buffer that exceeds
the 192 KB AIV budget. The batched program tiles this exact InCore over
the batch axis via `pl.range(batch)` with per-element GM slicing.

### Setup

Ascend 910B2 (24 cube cores, 64 GB HBM, FP32). Pushed kernel commit
`eaf6eb1`. Inputs pre-uploaded to NPU memory (per-call wall = launch +
execute only, no PCIe transfer). 3 warmup + 10 measured iters per shape;
median reported. Tools: msprof (`--task-time=on --aic-metrics=PipeUtilization`)
and PMU (`enable_pmu=2`).

### 1. Per-launch breakdown — host vs NPU (msprof)

For every shape, each launch decomposes into **NPU AIC kernel time + host-side dispatch gap**.

| B   | NPU dur (ms) | host gap (ms) | total wall (ms) | per-matrix |
|----:|---:|---:|---:|---:|
| 1   | 4.12 | 8.58 | 12.70 | 12 696 µs |
| 4   | 3.56 | 8.53 | 12.09 |  3 023 µs |
| 16  | 3.56 | 9.00 | 12.57 |    785 µs |
| 64  | 3.58 | 8.57 | 12.15 |    190 µs |
| 128 | 3.80 | 8.58 | 12.38 |     97 µs |
| 632 | 4.73 | 8.65 | 13.37 |     21 µs |

- **Host gap is flat at ~8.5 ms regardless of B** — AICPU dispatch + sync
  round-trip in `simpler.task_interface.run_prepared` (C++ binding;
  cProfile confirms 99.5% of host time is here, Python overhead is 0.5%).
- **NPU duration starts at 4.1 ms even for B=1** — fixed kernel floor; only
  grows to 4.7 ms at B=632 (15% more NPU for 632× more matrices).
- **Per-matrix cost amortizes 600× from B=1 → B=632** for only 5% more wall.

### 2. NPU engine utilization (PMU per-core averages)

| shape | cube_busy | mte1_busy | mte2_busy | scalar_busy | (100 − cube) |
|---|---:|---:|---:|---:|---:|
| n=128 B=1   | 49.5 % | 20.2 % | 19.2 % | 4.0 % | 50.5 % |
| n=128 B=4   | 47.3 % | 19.2 % | 22.5 % | 3.7 % | 52.7 % |
| n=128 B=16  | 46.6 % | 19.0 % | 23.2 % | 3.6 % | 53.4 % |
| n=128 B=64  | 53.5 % | 21.8 % | 18.7 % | 2.8 % | 46.5 % |
| n=128 B=128 | 55.0 % | 22.4 % | 17.6 % | 2.3 % | 45.0 % |
| n=64  B=1   | 22.8 % | 20.4 % | 24.3 % | 13.7 % | 77.2 % |
| n=64  B=64  | 25.9 % | 23.2 % | 25.3 % | 10.7 % | 74.1 % |

- **Cube engine 49–55% busy across n=128 cases** — best in class compared
  to canonical pypto-lib examples (gemm.py 15.6%, rms_norm.py 12.2%,
  softmax.py 18.4% under the same tools), i.e. **~3× more cube-active
  than the reference set**.
- **Memory pipeline (mte1 + mte2) stalls account for ~40% of NPU cycles**.
  This is the unavoidable cost of the L0C→GM→L1 round-trip per doubling
  iteration — Acc→Mat tile-move is unsupported on a2/a3 so X / Y cannot
  stay in L0 across iters.
- **n=64 drops to ~23% cube** — the [64,64] tile underfills the MAC unit
  and overhead ratios dominate; n=128 is the right operating point.

### 3. Per-engine work ratios (msprof `aic_*_ratio`)

| B   | aic_mac_ratio | aic_mte1_ratio | aic_mte2_ratio |
|----:|---:|---:|---:|
| 1   | 0.000 | 0.000 | 0.000 |
| 4   | 0.000 | 0.000 | 0.000 |
| 16  | 0.001 | 0.001 | 0.001 |
| 64  | 0.006 | 0.002 | 0.002 |
| 128 | 0.012 | 0.005 | 0.003 |
| 632 | 0.053 | 0.021 | 0.014 |

- `aic_mac_ratio` is the fraction of **AIC total cycles** spent actually
  in the MAC unit. Even at B=632 only **5.3% of cube cycles are matrix
  multiplication** — the kernel is dispatch- and memory-bound, not
  compute-bound. The [128,128]×[128,128] matmul is too small to amortize
  per-task scheduling cost; only batching brings ratios up at all.

### 4. Multi-core dispatch

| B | cube_utilization (% of 2400) | active cores ≈ |
|---:|---:|---:|
| 1   | 309 | 3.1 / 24 |
| 4   | 344 | 3.4 / 24 |
| 16  | 358 | 3.6 / 24 |
| 64  | 340 | 3.4 / 24 |
| 128 | 324 | 3.2 / 24 |
| 632 | 281 | 2.8 / 24 |

- **~3 of 24 cube cores active at every B**, set by `pl.matmul` parallelism
  inside one matrix inversion; **the outer `pl.range(batch)` loop dispatches
  serially on the AICPU lane** and does not fan out across cores.
- 87% of cube capacity is idle even at large B. Multi-core batch dispatch
  (e.g. `pl.spmd` over the batch axis) would unlock this — both refactor
  attempts in this branch compile but produce wrong output; needs
  IR-level investigation.

### 5. Buffer occupancy at n=128 (FP32)

| Cache | Capacity | Our usage | Headroom |
|---|---:|---:|---:|
| L0A (Left)  | 64 KB  | 64 KB   ([128,128] tile) | 0 — saturated |
| L0B (Right) | 64 KB  | 64 KB   ([128,128] tile) | 0 — saturated |
| L0C (Acc)   | 128 KB | 64 KB   ([128,128] tile) | 64 KB |
| L1 Mat      | ≥256 KB | 128 KB (A_l1 + I_l1) | comfortable |

- **L0A/L0B fill exactly at n=128**, which is why n>128 is rejected
  without K-blocking — operand tile would overflow.
- L0C has 64 KB headroom; would fit up to n=181 if K-blocking shrunk the
  operand tiles. A K-blocked variant was prototyped on a side branch
  (n=160 verified end-to-end, ~99 µs/matrix at B=128) but is out of scope
  for this PR.

### 6. Where the time goes — summary

```
Per-launch wall = 8.5 ms host gap + 3.5–4.7 ms NPU
                  ─────────────                  ─────────
                  64–71% of wall                  29–36% of wall
                  simpler-side                    kernel-side
                  (async submit)                  (memory + dispatch)
```

The kernel is at its dispatch-amortization sweet spot for the canonical
Qwen3-Next GDN workload (chunk=128). Per-matrix cost is **21 µs at B=632**.
Pushing per-matrix lower requires either (a) reducing the 8.5 ms host gap
in simpler, or (b) lighting up the 21 idle cube cores via parallel batch
dispatch — both outside the kernel-source scope of this PR.

### Known limits

- **n > 128** — `[n,n]` operand tile overflows L0A/L0B (64 KB). Lift
  requires K-blocked matmul inside the doubling loop.
- **B ≥ 15 in the GDN dispatch shape** (n_total = B × H × ⌈T/chunk⌉ ≥ 4740
  sequential tasks per launch) hits PTO2 ring-buffer overflow (AICPU
  507018). CI workaround bumps `PTO2_RING_TASK_WINDOW` etc.; without it,
  the practical ceiling is **B=10 → 51.8 M numel / launch**.

PR #372, opened 2026-05-25
refactor: drop deprecated pl.auto_chunk / chunked_loop_optimizer

lyfne123:
pypto#1504 removed pl.at(optimization=, split=) and the chunked_loop_optimizer sentinel, and deprecated pl.auto_chunk. Migrate all 22 kernels to explicit chunk loops: outer pl.parallel over chunk stride, pl.at inside, inner pl.range over the original step. chunk= values and name_hint preserved; pl.split kept where present.


lyfne123:
Superseded by #373 (migrate to optimizations=[pl.auto_chunk]).

PR #473, opened 2026-06-08
perf(dsv4 moe): pipeline gate/up K-loop + parallelize scatters

lwDavid:
## Summary

Two stacked latency optimizations for the DeepSeek-V4 decode MoE step (`models/deepseek/v4/moe.py`, single-card EP, FLASH config), measured on real a2a3 hardware with fixed seed and interleaved baseline/optimized pairs:

| | span (one decode step) | sum-busy (aic+aiv) |
|---|---|---|
| baseline | 1616 µs | 47.0 ms |
| + commit 1 (K-loop pipeline) | 1514 µs (−6.4 %) | 43.4 ms (−9.7 %) |
| + commit 2 (parallel scatters) | **1028 µs (−36.4 % total)** | 43.2 ms (unchanged) |

## Commit 1 — pipeline the gate/up K-loop

- `expert_routed`: run the gate/up K-accumulation under `pl.pipeline(stage=2)` (first iter peeled to seed the accumulator, avoiding pypto#1540), so each K-tile's weight MTE2 load overlaps the previous tile's cube compute. PMU shows this cube is MTE2-bound (~80 %); INT32 accumulation order is unchanged → bit-identical.
- The w2 K-loop stays serial: it is scalar-bound (~97 % PMU), pipelining buys nothing.
- `dispatch`: vectorize the tail zero-init (3 `pl.full` instead of 9216 scalar writes).

## Commit 2 — parallelize the two serial scatters (40 % of span)

`dispatch` (235 µs) and `combine_scatter` (380 µs) were single-core scalar kernels on the critical path. Both are now a 16-block `pl.spmd` over local experts, with block e owning exactly its expert's rows/column — no cross-block alias, zero→overwrite order preserved in-block, identical writers/values per slot:

- `combine_scatter`: block e zero-inits `routed_y_buf[:, e, :]` (strided 16-row fills) then scatters its recv rows. Wall 379.5 → 37.7 µs.
- `dispatch`: block e scans all T×TOPK routes, packing its own rows in scan order. Expert counts move to a tiny concurrent serial `dispatch_count` kernel (~14 µs, off critical path) because `recv_expert_count` `[16,1]` spans a single 64B cache line — 16 cores scalar-writing one line lose updates.
- `moe.py`: optional `--seed` for reproducible benchmarking (default keeps the original random behavior).

## Validation

- a2a3 device golden: 12/12 interleaved + 8/8 standalone rounds PASS (`ratio_reldiff(0.01, 0.05)`)
- a2a3sim: hash route, sort route (`--layer-id 5`), no-seed, plus `dispatch.py` / `combine.py` standalone goldens — all PASS
- `tests/`: 136/136 pass; kernel signatures, EP path (`dispatch_ep` / `combine_ep`) and CLI untouched

Per-pair interleaved delta: span −32.66 % ± 1.71 %, sum-busy +0.09 % ± 0.56 % (6 pairs, same card/session).

### DIFF
--- a/models/deepseek/v4/combine.py
+++ b/models/deepseek/v4/combine.py
@@ -45,18 +45,26 @@ def combine(
     routed_y_buf_flat = pl.reshape(routed_y_buf, [T * N_LOCAL_EXPERTS, D])
     recv_y_flat = pl.reshape(recv_y, [N_LOCAL_EXPERTS * RECV_MAX, D])
 
-    with pl.at(level=pl.Level.CORE_GROUP, name_hint="combine_scatter"):
-        for rb in pl.range(T * N_LOCAL_EXPERTS // 16):
-            r0 = rb * 16
-            routed_y_buf_flat[r0:r0 + 16, :] = pl.full([16, D], dtype=pl.BF16, value=0.0)
+    # Token-major 2D view: row t, cols [e*D, (e+1)*D) hold routed_y_buf[t, e, :].
+    routed_y_buf_2d = pl.reshape(routed_y_buf, [T, N_LOCAL_EXPERTS * D])
 
-        for e in pl.range(N_LOCAL_EXPERTS):
-            n_rows = pl.cast(pl.read(recv_expert_count, [e, 0]), pl.INDEX)
-            for s in pl.range(n_rows):
-                i = e * RECV_MAX + s
-                t = pl.cast(pl.read(recv_token, [e, s]), pl.INDEX)
-                dst = t * N_LOCAL_EXPERTS + e
-                routed_y_buf_flat[dst:dst+1, :] = recv_y_flat[i:i+1, :]
+    # Per-expert scatter: block e owns exactly the (t, e) cells of routed_y_buf
+    # (rows dst % N_LOCAL_EXPERTS == e in the flat view), so blocks never alias
+    # each other. Zero-init + scatter stay in the same block, preserving the
+    # zero -> overwrite ordering per cell that the duplicate-token (hash route)
+    # semantics require. ZERO_ROWS chunks keep each strided fill inside UB.
+    ZERO_ROWS = 16
+    for e in pl.spmd(N_LOCAL_EXPERTS, name_hint="combine_scatter"):
+        for z0 in pl.range(0, T, ZERO_ROWS):
+            routed_y_buf_2d[z0 : z0 + ZERO_ROWS, e * D : (e + 1) * D] = pl.full(
+                [ZERO_ROWS, D], dtype=pl.BF16, value=0.0
+            )
+        n_rows = pl.cast(pl.read(recv_expert_count, [e, 0]), pl.INDEX)
+        for s in pl.range(n_rows):
+            i = e * RECV_MAX + s
+            t = pl.cast(pl.read(recv_token, [e, s]), pl.INDEX)
+            dst = t * N_LOCAL_EXPERTS + e
+            routed_y_buf_flat[dst:dst+1, :] = recv_y_flat[i:i+1, :]
 
     for tb in pl.spmd(T // 4, name_hint="combine_reduce"):
         for tt in pl.range(4):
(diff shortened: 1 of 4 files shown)


lwDavid:
Rescoped to **dispatch-only**. The earlier `expert_routed` spmd-fold change has been removed: on re-measurement it did not produce a real end-to-end speedup — it only repartitions the same matmul work into fewer/longer InCore kernels (total core-busy time is unchanged), and on the CPU simulator it actually *increased* the modeled span ~15% from scheduling gaps; the earlier on-device "gain" was within run-to-run noise. That same fold (coarser, longer-running kernels) was also what tripped the x86_64 sim scheduler watchdog described above. With it removed, the PR is just the `dispatch` zero-init vectorization, which passes the sim in ~4 s.

lwDavid:
### Note on the `sim (a5sim)` check (timeout)

Everything is green except `sim (a5sim)`:
- ✅ `a2a3` (real Ascend 910B/C hardware), ✅ `sim (a2a3sim)`, ✅ pre-commit / unit-tests / detect-changes / CodeRabbit.
- ❌ `sim (a5sim)` — it times out at the 30-min job limit on the runtime cold-path scheduler watchdog tracked in **hw-native-sys/simpler#1014** (`scheduler_cold_path.cpp` core-deinit-timeout → `aicpu_execute` abort).

This is a simulator-scheduler-timing artifact, **not** a correctness or hardware problem:
- **Numerically identical** — the K-loop's INT32 accumulation order is unchanged; `pl.pipeline` only overlaps the weight load with compute. Golden passes everywhere, including on **real a2a3 hardware** and on **`a2a3sim`**.
- a5 = Ascend 950, **not** the chip this decode MoE targets (a2a3 / 910B-C).
- The watchdog is more sensitive on the a5 sim than a2a3sim for the same kernel: a2a3sim completes (~10 min), a5sim crosses the idle-iteration threshold and is cancelled. Data point added to simpler#1014.
- Precedent: prior DeepSeek-V4 PRs (#450, #463) merged with `sim` CANCELLED/FAILURE on these heavy decode kernels.

lwDavid:
### L2 swimlane — before / after (`--enable-l2-swimlane`, same seed)

Per-stage timeline on the same time scale (one decode step). `expert_routed` (the pipelined gate/up cube) and `dispatch` shrink, pulling everything after them earlier so the step finishes sooner:

```
BASELINE   (span 1585 us)
    |....:....|....:....|....:....|....:....|....:....|....:....|....:....|....:....|
1.hc_pre   ########
2.gate             #####
3.shared                 ########
4.dispatch                ##############
5.routed                                ################################
6.combine                                                               ##############
7.hc_post                                                                          ##

OPTIMIZED  (span 1508 us)        <- same scale; timeline ends earlier
1.hc_pre   #######
2.gate             #####
3.shared                #######
4.dispatch               ############        (dispatch -32 us: vectorized zero-init)
5.routed                              #########################   (routed -41 us: K-loop pipelined)
6.combine                                                          ####################
7.hc_post                                                                       ##
```

The merged Perfetto traces (per-core AICPU/AIC/AIV lanes) were captured with `--enable-l2-swimlane`; on the AIC lane the `exp_gate_up_aic` task is visibly shorter because the weight MTE2 load now overlaps the cube compute (PMU idle 20% → 12%).

lwDavid:
### Round 2 — interleaved benchmark for the parallel scatter commit (a2a3, `--seed 1234`)

6 baseline↔optimized pairs on the same card/session (per-pair delta cancels drift), each via `--runtime-dir` reuse:

```
pair  base span   opt span    Δspan     Δbusy
  1    1519.2     1025.7    −32.48 %   −0.28 %
  2    1529.2      999.6    −34.63 %   +0.49 %
  3    1539.1     1024.3    −33.45 %   +0.56 %
  4    1566.4     1035.5    −33.89 %   −0.87 %
  5    1515.6     1061.6    −29.96 %   +0.39 %
  6    1491.6     1021.3    −31.53 %   +0.22 %

mean Δspan = −32.66 % ± 1.71      mean Δbusy = +0.09 % ± 0.56
```

Critical-path kernels (wall, representative round):

```
                  before              after
dispatch          235.1 µs (1 core)   38.1 µs (16 blocks) + dispatch_count 13.7 µs (concurrent)
combine_scatter   379.5 µs (1 core)   37.7 µs (16 blocks)
```

All 12 interleaved runs + 8 standalone optimized rounds passed the on-device golden. Note the first device run of this approach failed golden (95 % mismatch) on real hardware while a2a3sim passed: real-parallel exposes both `pl.assemble`'s SSA chain (downstream sees one block's rows) and the 64B cache-line collision on the `[16,1]` count column — both worked around as described in the commit/code comments.

PR #503, opened 2026-06-11
[feature] add TurboQuant INT4 KV cache for Qwen3-14B

sunghajung6688:
关联issue: https://github.com/hw-native-sys/pypto-lib/issues/544

The dequant extraction refactoring (extracting `turboquant_kv_dequant_chunk`
from inline dequant logic duplicated across prefill and decode) was validated
on Ascend A2A3 NPU device 2 on 2026-06-16.

**Comparison baseline**: the shared `turboquant_kv_dequant_chunk` produces
bit-identical results to the original per-site inline dequant code (the
gather → renormalize → scale → unrotate sequence was not changed, only
factored into a single call site).

**Hardware**: A2A3 NPU (device 2)

| Test | compile | golden | runtime | result |
|------|---------|--------|---------|--------|
| `qwen3_14b_decode_tq.py` | 1.6s | 81.6s | 10.7s | `out` PASS (pass_rate ≥ 0.98) |
| `qwen3_14b_prefill_tq.py` | 2.1s | 26.1s | 7.8s | `out` PASS |

Both prefill and decode pass the full compile → golden → runtime → validate
pipeline, confirming the refactored dequant function is functionally
equivalent to the original duplicated code.

### DIFF
--- a/models/qwen3/14b/qwen3_14b_decode_tq.py
+++ b/models/qwen3/14b/qwen3_14b_decode_tq.py
@@ -1,0 +1,1301 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


bumble0918:
代码过长，按层次拆分commit；能复用的尽量复用

PR #564, opened 2026-06-20
feat(deepseek/v4): gather-free split-half RoPE for the decode path (draft, decode-only)

Hzfengsy:
## Summary

**Draft / direction proposal — decode-only, not yet validated by the serving system.**

Converts the DeepSeek-V4 **decode** RoPE path from the interleaved (GPT-J, gather-based) layout to **split-half** (GPT-NeoX, gather-free). The rotation partner of lane `k` becomes lane `k+HALF` — a contiguous `lo=[:HALF]` / `hi=[HALF:]` slice instead of a `j^1` swap-gather.

## What

- **Forward** (`qkv_proj_rope` q+kv, `decode_compressor_ratio{4,128}`): `out_lo = x_lo*cos − x_hi*sin`, `out_hi = x_lo*sin + x_hi*cos`. The compressors fold the per-column gamma per-half (gamma does not commute with the rotation).
- **Inverse** (`decode_sparse_attn{,_hca,_swa}`): `out_lo = x_lo*cos + x_hi*sin`, `out_hi = x_hi*cos − x_lo*sin` (the conjugate of forward).
- **Callers** (`decode_attention_{hca,swa,csa}`) feed half-width FP32 `rope_cos_half` / `rope_sin_half`; `decode_layer_dp_ep` drops the interleaved-table plumbing.
- Removes the in-kernel `rope_cs` dup-gather pre-pass and all `VGATHER`s from the RoPE kernels. Net −10 lines.

## Why

The interleaved layout forces a per-element gather (`j^1` swap + `j>>1` dup) that on-device profiling showed to be the dominant RoPE cost (an earlier mask-pattern alternative measured **+61%**, confirming the gather — not the arithmetic — is the bottleneck). Split-half makes every rotation a contiguous-slice + plain FMAs, with no cross-lane op.

## Validation (a2a3, on-device)

- **Bit-exact:** HCA standalone + composed + `--layer-id 3`; SWA standalone + composed + `--layer-id 0`; CSA standalone + composed; compressors 128 + 4 standalone. The tight standalone `ratio_allclose(1e-4, 1/128)` checks discriminate sign errors (a negative-control flipped sign produced 1.59 max-abs error).
- **L2 swimlane** (HCA attention module): rope compute 2970 → 877 µs (**−70.5%**), module wall-clock 1821 → 1645 µs (**−9.6%**). Per-kernel: `q_head_rope_fused` −77%, inverse `rope` −71%, `kv_rope_fused` −69%, `rmsnorm_rope` −38%, `rope_cs` eliminated. Non-rope kernels within ±2% noise.

## Caveats (why this is a draft)

- **Decode-only.** `qkv_proj_rope` is shared with prefill; prefill's own sparse_attn / compressors stay interleaved, so prefill is **latently half-converted** — its kernel==golden tests still pass, but the output is wrong vs the true model. Prefill conversion (~10 files, same pattern) is a tracked follow-up.
- **Deployment weight-permutation.** Real checkpoints need an offline permutation of the q-proj / k_pe / `wo_a` rope columns (interleaved → split-half) to stay bit-identical to the trained model. Synthetic tests need none; this is documented but not included here.
- **Pre-existing CSA-layer fault.** `decode_layer_dp_ep --layer-id 10` (CSA, 2-device) fails with AICore `507018` — but **`main` fails identically at the same point**, so it is pre-existing and unrelated to this change.
- **Not yet validated end-to-end by the serving system.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/models/deepseek/v4/decode_attention_csa.py
+++ b/models/deepseek/v4/decode_attention_csa.py
@@ -185,6 +185,11 @@ def attention_csa(
 
     rope_cos_t = pl.create_tensor([T, ROPE_HEAD_DIM], dtype=pl.BF16)
     rope_sin_t = pl.create_tensor([T, ROPE_HEAD_DIM], dtype=pl.BF16)
+    # Half-width unsigned inverse-RoPE snapshots for the split-half sparse_attn: the first
+    # HALF columns of the per-token cos/sin (one value per frequency), cast BF16->FP32 once per
+    # token so sparse_attn's per-head rope loop rotates with no in-loop cast and no gather.
+    rope_cos_half_t = pl.create_tensor([T, HALF_ROPE], dtype=pl.FP32)
+    rope_sin_half_t = pl.create_tensor([T, HALF_ROPE], dtype=pl.FP32)
     step_cos = pl.create_tensor([B, HALF_ROPE], dtype=pl.FP32)
     step_sin = pl.create_tensor([B, HALF_ROPE], dtype=pl.FP32)
     with pl.at(level=pl.Level.CORE_GROUP, name_hint="csa_rope_step"):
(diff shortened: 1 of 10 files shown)


zhangqi-chen:
Closing as stale: draft, conflicting, no activity since 2026-06-22 (51 days) and no human review. `models/` was flattened in #886 (2026-08-04) and the DSv4 decode RoPE path has moved on since. Please reopen or open a fresh PR against `main` if you want to revisit gather-free split-half RoPE.

PR #586, opened 2026-06-23
Add: vector (AIV) row_sum variant for qwen3-14b K/V projection

Inspiron-st:
## Summary
- Add env toggles `K_PROJ_ON_AIV` / `V_PROJ_ON_AIV` that run the qwen3-14b K/V projection on the VECTOR (AIV) unit as a dot-product / row_sum GEMM instead of the default cube matmul.
- Each wk/wv K-block is staged to UB and transposed on-chip so every output column becomes a contiguous row (avoids the strided GM column loads the AIV TLOAD rejects); operands upcast to FP32 to match the cube path's bf16-in / fp32-accumulate.
- K uses its own `K_RS_*` tiling: qk_norm consumes k_proj per 512-wide N-tile, so each AIV task is fanned into that N-tile's dependency slots.
- Default behavior is unchanged (cube); both variants validated on device against the torch golden (ratio_allclose 3e-3, <=2% outliers).

## Related Issues
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ Thumbs.db
 
 # Project-specific
 KNOWN_ISSUES.md
+
+# msprof
+PROF_*
--- a/models/qwen3/14b/qkv_proj_standalone.py
+++ b/models/qwen3/14b/qkv_proj_standalone.py
@@ -1,0 +1,357 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


zhangqi-chen:
Closing as stale — the target layout changed underneath this PR, and it never got a human review.

The new file lands at `models/qwen3/14b/qkv_proj_standalone.py`. That directory no longer exists: #886 flattened `models/` to one directory per model build on 2026-08-04, so the equivalent path today is `models/qwen3_14b/`. Merging as-is would recreate the old nested layout.

To be straight about it: this PR sat for over two months with only bot comments (gemini-code-assist, coderabbitai) and no human review. That is on us, not on the contribution, and the AIV `row_sum` variant for the K/V projection is a reasonable thing to want measured.

If you are still interested, a fresh PR placing the file in `models/qwen3_14b/` — with a wall-time comparison against the current projection path on a2a3 — would get looked at. Sorry for the wait, and thanks for the contribution.

PR #628, opened 2026-06-27
perf(dsv4 compressor): transpose kv_score_proj weights + b_trans (DN2ZN load, -7% wall)

wangqin1723-max:
## What

Store the kv/gate projection weights transposed (`[OUT_DIM, D]` instead of `[D, OUT_DIM]`) and consume them via `pl.matmul(..., b_trans=True)` in `decode_compressor_ratio4`'s `kv_score_proj`. The `decode_attention_csa` caller is updated to match (`cmp_wkv` / `cmp_wgate` now `[MAIN_OUT_DIM, D]`).

## Why

`kv_score_proj` is the compressor's critical-path scope and is **MTE2-load-bound**. The cube needs its B operand in NZ fractal layout, but the weights live in GM as ND, so every decode step re-converts them with an `ND2NZ` load (`MOV_OUT_TO_L1_MULTI_ND2NZ`).

That conversion is **transaction-bound, not bandwidth-bound**: building each 16×16 fractal block from row-major `[K, N]` issues many short strided bursts (~21 GB/s/lane observed, well under the HBM ceiling). The 5 MTE2 sub-queues run at only ~3.6/5 average concurrency, capped by the strided access pattern — and deeper pipelining can't help (stage=3 overflows the 512 KB Mat/L1 buffer).

Storing the weight as `[N, K]` and loading with `b_trans=True` turns it into a **DN2ZN**: each `[OUT_TILE, K_TILE]` row is K-contiguous, so the source reads are long contiguous bursts instead of many short ones. The fractal conversion still happens, but with far fewer transactions.

> Note: declaring the weight input as `pl.NZ` to skip the conversion entirely does **not** work — the layout annotation is parser-accepted but dropped by codegen (the GM `make_tensor_view` is hardcoded `nd`, generated `.pto` is byte-identical). `b_trans` is the cheapest lever that actually recovers the transaction overhead today.

## Result (rebased on current main, which includes #624 widen)

**Standalone `decode_compressor_ratio4`** — same-session A/B, clean 2-run (a2a3):

| metric | ND2NZ | b_trans (DN2ZN) | Δ |
|---|---|---|---|
| kv_score_proj busy | 6456 us | 5479 us | **−15.1%** |
| compressor total wall | 295 us | 270 us | **−8.3%** |

**Full `decode_attention_csa` graph** — median of 3 runs each (a2a3):

| metric | ND | b_trans | Δ |
|---|---|---|---|
| kv_score_proj busy | 1299 us | 912 us | **−29.8%** |
| graph total wall | 4304 us | 3750 us | **−12.9%** |

> Noise caveat: the CSA full-graph L2-swimlane union-span wall is session-sensitive (idle/dispatch gaps inflate it — absolute values ran ~2.7× higher this session than another). Single CSA runs are unreliable (one pair even showed +20%); the 3-run medians are reported because all b_trans runs land below all-but-one ND run, so the direction is robust even though the exact magnitude (≈ −10%, range −8…−15%) is not. The standalone −8.3% and the kv_score_proj busy reductions are the firm numbers.

## Cost

The weight transpose is a one-time host-side layout change at model load (free per-step). Callers must feed `wkv` / `wgate` as `[OUT_DIM, D]`.

## Validation (a2a3)

- standalone compressor: `kv` / `compress_state` / `cmp_kv_cache` — **PASS**
- `decode_attention_csa` full graph: `x_out` / `kv_cache` — **PASS**


zhangqi-chen:
Nice win, and the CSA chain (compressor_ratio4 / decode_attention_csa / decode_fwd / decode_layer) is internally consistent. One follow-up before this is end-to-end correct on real weights — please sync the HCA path the same way:

**Why it matters now**

The native checkpoint already stores `compressor.wkv.weight` / `wgate.weight` as `[OUT_DIM, D]` (verified on `dsv4-flash-w8a8`: `layers.N.attn.compressor.wkv.weight` is `[1024, 4096]`). So the `[OUT_DIM, D]` layout this PR adopts *is* the native layout — but `extract_weights_w8a8.py` currently `.t()`s it back to `[D, MAIN_OUT_DIM]` (lines 96-97), and that `cmp_wkv`/`cmp_wgate` extraction is **shared by both CSA and HCA**. After this PR, CSA wants native `[OUT_DIM, D]` while HCA still wants `[D, OUT_DIM]`, so the two consumers disagree and no single `extract_weights` layout is correct. CI does not catch this because the fixtures are random-init.

**Proposed follow-up (decode + prefill HCA, same transform as this PR)**

- `decode_compressor_ratio128.py`: `kv_score_proj` at L96-111 is structurally identical to ratio4 — store `wkv`/`wgate` as `[OUT_DIM, D]`, consume via `b_trans=True`, update the test/golden (`x @ wkv.t()`) and inits.
- `decode_attention_hca.py`: `cmp_wkv`/`cmp_wgate` `[D, MAIN_OUT_DIM]` -> `[MAIN_OUT_DIM, D]` (both `attention_hca` and `attention_hca_test`).
- `decode_fwd.py` / `decode_layer.py`: the `hca_cmp_wkv` / `hca_cmp_wgate` params + slices.
- prefill equivalents (`prefill_attention_hca.py`, prefill compressor, `prefill_fwd.py`, `prefill_layer.py`).
- `extract_weights_w8a8.py`: drop the `.t()` on `cmp_wkv`/`cmp_wgate` (and `inner_wkv` if the indexer compressor gets the same treatment) once both CSA and HCA expect native layout.

This keeps the extraction uniform, makes the real-weight path correct, and HCA gets the same MTE2 transaction win for free.

Side note: the CodeRabbit comment about `decode_fwd.py` not being updated is a false positive — its review scope only included 2 files; the `decode_fwd.py`/`decode_layer.py` shapes are already updated in this PR.

wangqin1723-max:
## Follow-up required for real-weight e2e: drop `.t()` in `extract_weights_w8a8.py`

This PR now propagates the `b_trans` transposed-weight layout (`[OUT_DIM, D]`) across **all** compressor consumers, so they are mutually consistent:

| consumer | layout |
|---|---|
| decode CSA (`decode_compressor_ratio4`) | `[OUT_DIM, D]` |
| decode HCA (`decode_compressor_ratio128`) | `[OUT_DIM, D]` |
| prefill CSA (`prefill_compressor_ratio4`) | `[OUT_DIM, D]` |
| prefill HCA (`prefill_compressor_ratio128`) | `[OUT_DIM, D]` |

The native checkpoint already stores `compressor.wkv.weight` / `wgate.weight` as `[OUT_DIM, D]` (verified `[1024, 4096]` on dsv4-flash-w8a8). `extract_weights_w8a8.py` (lines 96-97) currently `.t()`s them back to `[D, OUT_DIM]`, which was correct for the pre-`b_trans` kernels but is now **wrong** — it would feed `[D, OUT_DIM]` to kernels that expect `[OUT_DIM, D]`.

**CI does not catch this** because every standalone test uses random fixtures initialized in the new `[OUT_DIM, D]` shape; only the real-weight path (checkpoint → `extract_weights`) is affected.

### Action (in the checkout that owns `extract_weights_w8a8.py`)

Drop the `.t()` on `cmp_wkv` / `cmp_wgate` only:

```python
        out["cmp_wkv"]   = g(f"{cpfx}.wkv.weight").contiguous()    # [MAIN_OUT_DIM, D] native (kernels use b_trans=True)
        out["cmp_wgate"] = g(f"{cpfx}.wgate.weight").contiguous()
```

⚠️ Keep `.t()` on `inner_wkv` / `inner_wgate` (lines 106-107) — the indexer compressor is **not** converted in this PR.

After applying, run the dsv4-flash-w8a8 real-weight e2e to confirm correctness.

### Perf note (same-session L2 swimlane A/B, pypto #1866)

- `kv_score_proj` Avg Exec: CSA (ratio4) **−24.8%**, HCA (ratio128) **−28.9%** standalone (**−56%** inside the full HCA graph).
- compressor wall: ratio4 **−17%** (kv_score_proj-dominated); ratio128 **−3.7%** and full HCA graph ~flat — ratio128/HCA is `softmax_pool` + `qk_pv` bound, so the b_trans win there is **consistency/correctness**, not an end-to-end HCA perf lever.

wangqin1723-max:
[Uploading extract_weights_w8a8_drop_cmp_t.patch…]()

PR #663, opened 2026-07-01
Docs: add bump-toolchain-versions skill

chenshengxin2026:
## What

Adds `.claude/skills/bump-toolchain-versions/SKILL.md` — an SOP skill for
advancing the two external toolchain versions the CI pipelines depend on:

- **ptoas** — the assembler binary on the codegen -> device path
- **pto-isa** — the ISA header repo, pinned by a 40-char commit SHA

across the three repos (`simpler`, `pypto`, `pypto-lib`).

## Why

The bump procedure has a strict ownership model (one source of truth per
artifact) and a required propagation order. Capturing it as a skill keeps the
process reproducible and prevents editing derived values by mistake.

## Contents

- Ownership model: `simpler/pto_isa.pin` owns pto-isa; `pypto/toolchain/versions.env`
  owns ptoas; `pypto-lib` inherits both from pypto's `main` with no code edit
  under the current derive-from-pypto CI layout.
- Propagation order: simpler -> pypto -> pypto-lib.
- Fork-and-PR flow, value-gathering steps with self-checks, verification/monitoring,
  and a checklist.
- Known CI gotchas (cold-cache `.tmp` race on the first run of a new version,
  onboard flakes) so they are not misdiagnosed as a bad bump.

Documentation-only change under `.claude/skills/`; no runtime or kernel code is affected.

### DIFF
--- a/.claude/skills/bump-toolchain-versions/SKILL.md
+++ b/.claude/skills/bump-toolchain-versions/SKILL.md
@@ -1,0 +1,330 @@
+---
+name: bump-toolchain-versions
+description: Update the ptoas and/or pto-isa toolchain versions that CI depends on across the three repos (simpler, pypto, pypto-lib), then open the PRs. First confirms the target ptoas version with the user (latest release vs. a user-specified tag) rather than assuming latest. Use when the user wants to bump ptoas to a new PTOAS release, move pto-isa to a newer commit, or sync toolchain versions across simpler/pypto/pypto-lib.
+---
+
+# Bump CI Toolchain Versions (ptoas / pto-isa) across simpler, pypto, pypto-lib
+
+SOP for advancing the two external toolchain versions the CI pipelines depend on:
+
+- **`ptoas`** — the assembler binary on the codegen → device path
+  (`github.com/hw-native-sys/PTOAS` releases).
+- **`pto-isa`** — the ISA header repo (`github.com/hw-native-sys/pto-isa`),
+  pinned by a 40-char commit SHA.
+
+## Ownership model — who owns what, who inherits
+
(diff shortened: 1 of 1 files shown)


chenshengxin2026:
Updated the skill to fix a correctness issue in the pto-isa propagation model (synced from the same fix applied to the simpler/pypto copy).

**Problem:** the SOP treated "bump the `runtime/` submodule pointer" as a mechanical step of a pto-isa bump. But pypto has **no independent pto-isa knob** — the `toolchain` job derives `PTO_ISA_COMMIT` *only* from `runtime/pto_isa.pin` (the simpler commit the submodule points at), and the guard forbids pinning it elsewhere. The submodule points at a *whole* simpler commit, so moving pypto's pto-isa necessarily moves the **runtime** by however many commits it lags simpler main. When that lag is non-trivial, it's really a runtime version bump (adaptation-prone) — the pypto team's routine `chore(runtime)` work, not a toolchain step.

**Changes (content only; the `$GH_USER`/`$GIT_NAME`/`$GIT_EMAIL` de-hardcoding in this PR is untouched):**
- Ownership model: added a ⚠️ consequence spelling out the coupling.
- Propagation order + Step 1 note: pto-isa version bump = the **simpler pin PR**; pypto pickup is a separate runtime-bump decision, not automatic.
- Step 2b rewritten to **gate on the runtime lag** — STOP and defer to the pypto team if the lag is non-trivial; only proceed (own PR, target simpler-main HEAD) if trivial and explicitly wanted. Uses `git update-index --cacheinfo` to move only the gitlink.
- Checklist updated to match.

PR scope is still a single file (`.claude/skills/bump-toolchain-versions/SKILL.md`); commit amended to keep the "add skill" history clean.

PR #688, opened 2026-07-03
Enable DeepSeek V4 decode 8k metadata support

sjduan:
## Summary
- Raise DeepSeek V4 Flash decode capacity to 16k positions and 32 compressed KV blocks so the 8k prompt + 512 decode target can be represented.
- Add shared decode metadata lowering for long absolute positions, block tables, physical slot mappings, compressor state mappings, and single-commit speculative boundary masking.
- Update SWA/HCA/CSA decode attention and ratio4/ratio128/indexer compressors to consume INT64 physical slot mappings with >=0 no-write guards.
- Widen logical compressor state block tables for MAX_SEQ_LEN while keeping the physical rolling state buffers bounded.
- Update CSA indexer topk to cover 4096 scores by sorting two 2048 halves and merging the per-half top candidates.
- Update decode_fwd metadata lowering for long positions and single-commit visible length so repeated one-token decode calls can reach committed position 8703.

## Scope Notes
- This is the minimized upstream diff for decode 8k enablement. Debug-only phase harnesses and metadata-case CLIs were removed from the final patch.
- No 512-step monolithic JIT is introduced. The intended serving path is repeated decode_fwd calls with Phase4 single-commit speculative overwrite semantics.
- CSA standalone x_out compare uses diff_thd=4e-3 to cover one BF16 step around unit-scale values at high positions; kv_cache compare remains strict.

## Tests
Local:
- python3 -m py_compile models/deepseek/v4/config.py models/deepseek/v4/decode_metadata.py models/deepseek/v4/decode_attention_swa.py models/deepseek/v4/decode_attention_hca.py models/deepseek/v4/decode_attention_csa.py models/deepseek/v4/decode_compressor_ratio4.py models/deepseek/v4/decode_compressor_ratio128.py models/deepseek/v4/decode_indexer_compressor.py models/deepseek/v4/decode_indexer.py models/deepseek/v4/decode_fwd.py
- git diff --check

Remote a2a3, EP-independent:
- decode_compressor_ratio4.py --start-pos 8702
- decode_compressor_ratio128.py --start-pos 8702
- decode_indexer.py --start-pos 16382
- decode_attention_swa.py --start-pos 8191
- decode_attention_hca.py --start-pos 8702
- decode_attention_csa.py default / --start-pos 8192 / --start-pos 8702

Remote a2a3, EP2:
- decode_layer.py --ep 2 --layer-id 0 --start-pos 8191
- decode_layer.py --ep 2 --layer-id 2 --start-pos 8702
- decode_layer.py --ep 2 --layer-id 3 --start-pos 8702
- decode_fwd.py --ep 2 --start-pos 8192 --num-tokens 8
- decode_fwd.py --ep 2 --start-pos 8702 --num-tokens 8

Remote a2a3, EP8 smoke:
- decode_fwd.py --ep 8 --start-pos 8702 --num-tokens 8 on devices 2,3,4,5,12,13,14,15

### DIFF
--- a/models/deepseek/v4/config.py
+++ b/models/deepseek/v4/config.py
@@ -174,7 +174,7 @@ FLASH = DeepSeekV4Config(
     hc_mult=4,
     hc_sinkhorn_iters=20,
     hc_eps=1e-6,
-    max_position_embeddings=8192,  # official 1M;
+    max_position_embeddings=16384,  # 8k prompt + 512 decode steps target; official 1M;
     rope_theta=10000.0,
     compress_rope_theta=160000.0,
     rope_factor=16.0,
@@ -257,7 +257,7 @@ LM_HEAD_TP_SIZE = 8
 # decode-batch-sized physical pool because serving slot ids address the shared
 # decode KV-cache layout even when a prefill kernel handles one request at a time.
 KV_ORI_MAX_BLOCKS = 1
-KV_CMP_MAX_BLOCKS = 8
+KV_CMP_MAX_BLOCKS = 32
 IDX_CACHE_MAX_BLOCKS = 64
 DECODE_ORI_BLOCK_NUM = DECODE_BATCH * KV_ORI_MAX_BLOCKS
 DECODE_CMP_BLOCK_NUM = DECODE_BATCH * KV_CMP_MAX_BLOCKS
--- a/models/deepseek/v4/decode_attention_csa.py
+++ b/models/deepseek/v4/decode_attention_csa.py
@@ -84,13 +84,15 @@ COFF = 1 + int(OVERLAP)
 MAIN_OUT_DIM = COFF * HEAD_DIM
 MAIN_STATE_DIM = 2 * MAIN_OUT_DIM
 MAIN_STATE_BLOCK_SIZE = C4A_COMPRESSOR_BLOCK_SIZE
-MAIN_STATE_MAX_BLOCKS = 65
-MAIN_STATE_BLOCK_NUM = B * MAIN_STATE_MAX_BLOCKS
+MAIN_STATE_PHYSICAL_BLOCKS = 65
+MAIN_STATE_MAX_BLOCKS = (MAX_SEQ_LEN + MAIN_STATE_BLOCK_SIZE - 1) // MAIN_STATE_BLOCK_SIZE
+MAIN_STATE_BLOCK_NUM = B * MAIN_STATE_PHYSICAL_BLOCKS
 INNER_OUT_DIM = COFF * IDX_HEAD_DIM
 INNER_STATE_DIM = 2 * INNER_OUT_DIM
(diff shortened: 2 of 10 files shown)


sjduan:
Rebased onto `origin/main` / `upstream/main` at `bbd7d93` and refreshed the branch to `3bec99f`.

One small follow-up fix was included after rebase: `decode_indexer.py` test wrapper now keeps `idx_kv_cache` as a normal typed tensor parameter, matching the indexer core signature. With the newer frontend, the previous `pl.InOut[...]` annotation was lowered as an untyped parameter in the orchestration wrapper.

Validation after rebase:
- Local: `python3 -m py_compile` for decode metadata/indexer/fwd and related files: PASS
- Local: `git diff --check`: PASS
- Remote 2号机 container (`/home/sj/git/pypto-lib`, current toolchain):
  - `decode_indexer.py --start-pos 16382`: PASS
  - `decode_attention_csa.py --start-pos 8702`: PASS
  - `decode_attention_swa.py --start-pos 8191`: PASS
  - `decode_attention_hca.py --start-pos 8702`: PASS
  - `decode_layer.py --ep 2 --layer-id 2 --start-pos 8702`: PASS
  - `decode_fwd.py --ep 2 --start-pos 8702 --num-tokens 8`: PASS (runtime PASS; script has no golden compare)

Note: the older PyPTO remote machine currently fails latest `hc_pre` from upstream main with the split-aiv parser check, so I used the 2号机 container with the newer toolchain for the final validation.

sjduan:
Follow-up for the `decode_indexer.py` CI failure:

The previous workaround made `idx_kv_cache` a plain tensor too deep in the call chain, which allowed compile but broke the standalone output comparison: the rows written by the inner indexer compressor were not reliably reflected in the returned `idx_kv_cache`.

I updated the branch to keep the public `indexer_test` wrapper parameter as a normal typed tensor for the newer frontend, but restored `indexer(...)` core's `idx_kv_cache` parameter to `pl.InOut[...]`, which is the point where the cache is actually mutated and then read for scoring.

Validated on 2号机 container:
- `decode_indexer.py -p a2a3 -d 0`: PASS, including `idx_kv_cache`
- `decode_indexer.py -p a2a3 -d 0 --start-pos 16382`: PASS, including `idx_kv_cache`

sjduan:
Updated `decode_indexer.py` again to follow the baseline style exactly: both `indexer(...)` and `indexer_test(...)` now keep `idx_kv_cache` as `pl.InOut[...]`, and the test wrapper returns it directly.

The previous mixed style (`indexer_test` plain tensor + `indexer` InOut) was too clever and still allowed CI to miss the cache mutation in the returned output. This version matches the existing compressor/indexer standalone pattern.

Validated on 2号机 container:
- `python models/deepseek/v4/decode_indexer.py -p a2a3 -d 0`: PASS, including `idx_kv_cache`
- `python models/deepseek/v4/decode_indexer.py -p a2a3 -d 0 --start-pos 16382`: PASS, including `idx_kv_cache`

PR #742, opened 2026-07-09
fix(dsv4/moe): revert SPMD dispatch/combine — unbreaks ep8 prefill (507015)

zhangqi-chen:
## Summary

Real-weight ep8 `prefill_fwd` (43 layers, DeepSeek-V4-Flash W8A8, 8 cards) dies on `main` with an AICPU stream-sync timeout — 507015/507018, all 8 ranks fail to drain, and the first reporting card drifts (dev 9/10/11). A bisect pins it to #705, which SPMD-parallelized MoE `dispatch_push`/`dispatch_gather`; #736 then did the same to `combine`.

This restores `dispatch` and `combine` to their pre-#705 form — `dispatch_meta` / `dispatch_push` / `dispatch_gather` and `combine` each a single `pl.at(CORE_GROUP)` — dropping the SPMD grids, the separate `dispatch_wait` / `combine_wait` handshake tasks, and the now-unused `D_GATHER_TILE`. `moe.py` is the only file #705 and #736 touched.

**This is not a performance rollback.** At 197.82s the reverted tree is *faster* than either passing baseline; the other MoE work on `main` is kept and is a genuine win.

## Evidence

Bisect over `11270d33..c99ce1a8`. First bad commit is `199a6d4a` (#705); its parent `ea003140` (#733) passes, and the two are adjacent, so the attribution is exact rather than an interval.

| tree | runtime | result |
|---|---|---|
| `11270d33` (baseline) | 271.39s | PASS |
| `ea003140` (#733) | 329.42s | PASS |
| **`main` + this revert** | **197.82s** | **PASS** |
| `199a6d4a` (#705) | — | FAIL 507015 |
| `c99ce1a8` (`main`) | 684–834s | FAIL 507015/507018 |

`ea003140..c99ce1a8` also touches two other files inside `prefill_fwd`'s import closure — `expert_routed.py` (#737, w2 activation epilogue → spmd) and `hc_pre.py` (drops the `write_post` spmd grid and a GM round-trip). Both stay at `main` in the passing run above, so neither contributes; both are kept.

## Why it hangs

Device compute is not the bottleneck: `device_wall` is ~328 ms on the passing baseline, so the 271s → 834s growth is task-graph build and AICPU scheduling, not math.

At ep8 `N_LOCAL = 32`, so per MoE call the task count goes

```
#733:  dispatch_meta 1 + dispatch_push 1 + dispatch_gather 1 + combine 1 + shared_routed T
main:  dispatch_meta 1 + dispatch_push 32 + dispatch_wait 1 + dispatch_gather 32
       + combine_base 1 + combine 32 + combine_wait 1 + shared_routed T
```

132 → 228 per layer, roughly +4100 tasks over 43 layers, with 32-way fan-in/fan-out on every `grid → barrier → grid` edge.

Splitting by local expert also breaks the "push a peer, then notify that peer" locality that program order gave for free inside one CORE_GROUP task. The code says so itself:

> `dispatch_wait` … Kept separate from the SPMD gather below, **which cannot host the per-rank notify/wait**.

> `combine_wait` … since **a per-expert block spans all peers and cannot host a per-peer notify/wait**.

So #705 hoisted the handshake into its own task fenced on the whole grid, turning a pipelined per-peer handshake into a global barrier — 8 ranks × 43 layers × 2 sites.

## Why prefill and not decode

Every cost #705/#736 add scales with `T` (prefill 128, decode 8) and `N_LOCAL` (32 at ep8):

- `shared_routed = pl.spmd(T)` is 8 blocks on decode, 128 on prefill, so prefill's MoE task count is ~2× decode's before any of this.
- `dispatch_gather` sizes `D_GATHER_TILE` so `MAX_PER_SRC * tile ≤ 96 KB`, and `MAX_PER_SRC = T`. Decode (`8 × 7168 ≤ 96 K`) moves a whole row in one pass; prefill halves down to 448, needing `D / 448 = 16` load/store passes per (expert, source) — 32 blocks × 8 sources × 16 = 4096 tile ops per layer.
- Each `dispatch_push` block rescans all `T * TOPK` routes and acts only on its own expert: 768 scalar GM reads per block on prefill (×32 blocks) vs 48 on decode.

`main` is decode-tuned; prefill sits far closer to the ring limits and #705 pushed it over. Consistent with that, doubling `PTO2_RING_DEP_POOL` to 32768 did not help and ran slower still (834s), pointing at the task window / dep fan-in rather than dep-pool capacity.

## Follow-up

A structural fix that keeps the parallelism would split the grid by **destination rank** rather than local expert, so each block owns one peer and can host its own notify/wait — recovering the per-peer handshake and removing the global barrier. Left out of this PR, which is scoped to unbreaking `main`.

## Test

- `python models/deepseek/v4/_prefill_fwd_real_draft.py -p a2a3 --ep=8 -d {}` on cards 8–15, real W8A8 weights — PASS (197.82s runtime).
- `prefill_fwd` has no `golden_fn`, so this is a smoke run: it confirms the network runs end to end without the AICPU timeout, not numerical equivalence.

### DIFF
--- a/models/deepseek/v4/moe.py
+++ b/models/deepseek/v4/moe.py
@@ -73,25 +73,14 @@ AUX_W = 1
 IDX_PAD = 8  # INT32 route tile width; route rides a separate window from scale/w
              # (an FP32 tile can't hold it: INDEX->FP32 casts are unsupported).
 
-# The SPMD gather stages the payload through a [MAX_PER_SRC, D_GATHER_TILE] INT8
-# tile; bound it within the AICore UB. Decode (MAX_PER_SRC=T=8) keeps the whole row
-# (one 32 KB load); prefill (MAX_PER_SRC=128) tiles D so 128 * D_GATHER_TILE fits.
-# D is a power of two, so halving keeps D % D_GATHER_TILE == 0.
-_GATHER_X_UB = 96 * 1024
-D_GATHER_TILE = D
-while MAX_PER_SRC * D_GATHER_TILE > _GATHER_X_UB and D_GATHER_TILE > 1:
-    D_GATHER_TILE //= 2
-
 assert N_RANKS in _EP_CHOICES, f"--ep must be one of {_EP_CHOICES} (got {N_RANKS})"
 assert N_EXPERTS_GLOBAL == N_RANKS * N_LOCAL
 assert RECV_MAX == N_RANKS * MAX_PER_SRC
 
 
 # === Dispatch ================================================================
-# Three tasks: dispatch_meta (count routes, publish + barrier per-expert counts,
-# cumsum) and dispatch_push (SPMD (dst, loc_e) payload scatter) run independently
-# and overlap; dispatch_gather (deps on both) does the payload arrival handshake
-# then compacts each lane into the per-expert buffers.
+# Lane push, count publish, arrival wait, and cumsum gather run in one
+# pl.at(CORE_GROUP) so program order stays push -> notify -> wait -> gather.
 @pl.jit.inline
 def dispatch(
     indices: pl.Tensor[[T, TOPK], pl.INT32],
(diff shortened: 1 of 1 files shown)


zhangqi-chen:
## Validation on this branch

Real-weight ep8 `_prefill_fwd_real_draft.py` (43 layers, DeepSeek-V4-Flash W8A8, cards 8-15), `main` + this revert:

```
[RUN] compile done (24.13s)
[RUN] generate inputs done (450.67s)
[RUN] runtime done (197.82s)
[RUN] PASS (672.62s, validation skipped: no golden_fn or golden_data)
```

Full picture across the bisect:

| tree | runtime | result |
|---|---|---|
| `11270d33` (baseline) | 271.39s | PASS |
| `ea003140` (#733) | 329.42s | PASS |
| **`main` + this revert** | **197.82s** | **PASS** |
| `199a6d4a` (#705) | - | FAIL 507015 |
| `c99ce1a8` (`main`) | 684-834s | FAIL 507015/507018 |

Two things this pins down:

1. `moe.py`'s dispatch/combine is the **sole** cause. `ea003140 -> c99ce1a8` also touches two other files inside `prefill_fwd`'s import closure — `expert_routed.py` (#737, w2 activation epilogue -> spmd) and `hc_pre.py` (drops the `write_post` spmd grid + a GM round-trip). Both stay at `main` here and the run passes, so neither contributes.

2. This is **not** a performance rollback. At 197.82s the reverted tree is faster than either passing baseline (271s / 329s), because #737's epilogue spmd and hc_pre's grid removal are genuine wins that survive. Only the dispatch/combine SPMD grids are removed.

PR #785, opened 2026-07-15
perf(dsv4 hca decode): early-resolve gather_kv/merge_norm + L2 swimlane level flag

wangqin1723-max:
## Summary
- `decode_sparse_attn_hca.py`: add `allow_early_resolve` to `hca_gather_kv` and `merge_norm`, completing the producer sets for the already-flagged `qk_pv` / `proj_a_mm` consumers so they can pre-stage (early dispatch).
- `decode_attention_hca.py`: `--enable-l2-swimlane` now takes a granularity level (1=AICore timing, 2=+dispatch/fanout, 3=+sched phases, 4=+orch phases; bare flag = 4/full).

## Performance
`hca_gather_kv` is the main lever — it completes `qk_pv`'s producer set so `qk_pv` can pre-stage (its dispatch bubble → ~0) and stabilizes the HCA decode AICore time. `merge_norm` is neutral (`proj_a_mm` is an O_GROUPS parallel reduction — core-contention-bound rather than dispatch-fanin-bound); kept for producer-set completeness.

6-run AICore end-to-end on this config (a2a3, `--start-pos 8192`): **310.9 / 312.9 / 314.4 / 319.0 / 320.7 / 326.7 us** (median ~317, σ≈6).

## Validation
- golden **PASS** (kv_cache + x_out), a2a3 device, `--start-pos 8192`.

### DIFF
--- a/models/deepseek/v4/decode_attention_hca.py
+++ b/models/deepseek/v4/decode_attention_hca.py
@@ -657,7 +657,17 @@ if __name__ == "__main__":
     parser.add_argument("--start-pos", type=int, default=None,
                         help="Uniform fixture-only start_pos override for all batches; "
                              "default (unset) uses the canonical per-batch HCA set that includes the 8k point.")
-    parser.add_argument("--enable-l2-swimlane", type=int, nargs="?", const=1, default=0, choices=(0, 1, 2))
+    parser.add_argument(
+        "--enable-l2-swimlane",
+        nargs="?",
+        const=4,
+        default=0,
+        type=int,
+        metavar="PERF_LEVEL",
+        help="Enable L2 swimlane perf capture at the given granularity level. Bare flag "
+             "= level 4 (full). Levels: 1=AICore timing, 2=+dispatch/fanout, 3=+sched "
+             "phases, 4=+orch phases; 0 (default) disables.",
+    )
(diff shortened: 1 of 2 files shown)


zhangqi-chen:
Closing as stale — both halves of this PR are now obsolete against main `e058d747`.

1. **The swimlane flag half is superseded.** `--enable-l2-swimlane` was renamed repo-wide to `--enable-chip-swimlane` in #1012, and the level semantics this PR documents were reworked with it. The diff no longer applies to any argument parser in the tree.
2. **The `allow_early_resolve` hint on `hca_gather_kv` was not adopted, and the tree has since moved away from it.** #1042 removed `allow_early_resolve` from the DeepSeek-V4 decode and prefill paths after it was shown to be non-deterministic on a5 and perf-neutral (see #1043). Adding the hint back to a gather would need that evidence rebutted first.

On top of that, the target paths `models/deepseek/v4/` no longer exist: #886 flattened `models/` to one directory per model build on 2026-08-04, so this PR is `CONFLICTING` in a way rebasing cannot resolve.

Nothing here is a judgement on the underlying idea. If the early-resolve staging is worth revisiting on `models/deepseek_v4_flash_mtp/decode_sparse_attn_hca.py`, please open a fresh PR with before/after swimlanes on the current layout.

PR #795, opened 2026-07-18
feat(qwen3): add --decode-steps for autoregressive decode timing in decode_fwd

ChaoWao:
## Summary
`decode_fwd.py --validate-fwd` dispatched `decode_fwd` exactly once, so a
STRACE / L2-swimlane timing run only captured a **single** decode invocation —
not a steady-state batch decode.

Add **`--decode-steps N`**: run N autoregressive decode steps. Each step feeds
the previous step's sampled token back as input and **grows the context by one**
(`seq_lens += 1`, `slot_mapping` re-pointed to the new token's KV slot via
`_paged_block_table_slot_mapping`). The sweep starts at `MAX_SEQ - decode_steps`
and ends at `MAX_SEQ`, so the KV pool already sized for `MAX_SEQ`
(`PTO2_MANUAL_MAX_SEQ`) holds the whole run with no overflow (an `assert` guards
`decode_steps < MAX_SEQ`). Host-reference argmax check is skipped for `N > 1`
(perf-only). `N=1` (default) preserves the existing single-step behavior.

## Why
Enables a stable steady-state batch-decode timing sweep from one run, e.g. the
full-model batch-16 × 3338-context decode:
```bash
PTO2_MANUAL_MAX_SEQ=3358 python decode_fwd.py -p a2a3 -d 0 \
  --validate-fwd --fwd-layers 40 --max-seq --decode-steps 20
```
STRACE then reports the median device `Effective` over 20 real growing-context
decode invocations (measured ~37.7 ms/step for batch-16, 40L, ctx 3338→3358).

## Test
- `--decode-steps 20` on a2a3 (batch-16, 40L, ctx 3338→3358): grows without KV
  overflow, 20 STRACE decode invocations, run completes cleanly.
- Default `--decode-steps 1`: unchanged single-step validate path.

### DIFF
--- a/models/qwen3/14b/decode_fwd.py
+++ b/models/qwen3/14b/decode_fwd.py
@@ -1557,6 +1557,17 @@ if __name__ == "__main__":
                              "-> logits) against a host chain reference, instead of the default "
                              "single-layer golden test.")
     parser.add_argument("--fwd-layers", type=int, default=4, help="layer count N for --validate-fwd")
+    parser.add_argument("--decode-steps", type=int, default=1,
+                        help="under --validate-fwd, run this many decode steps for a device-cost "
+                             "timing sweep at growing context: each step feeds the previous step's "
+                             "sampled token back as input and grows the context by one (seq_lens += 1, "
+                             "slot_mapping advanced to the new token's KV slot). The context starts at "
+                             "MAX_SEQ (PTO2_MANUAL_MAX_SEQ) and grows to MAX_SEQ + decode_steps - 1; the "
+                             "paged KV pool (block_table / slot_mapping / kc / vc) is enlarged by "
+                             "decode_steps tokens up front so the growing seq_lens never overflow it. "
+                             "Read the device `Effective` span (kc/vc are re-uploaded per step, so it "
+                             "measures decode cost at growing context, not a correct generation). The "
+                             "host-ref argmax check is skipped for N > 1.")
     parser.add_argument("--save-data", action="store_true", default=False,
                         help="persist inputs + golden for replay (off: large fixtures)")
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


ChaoWao:
Thanks for the reviews — addressed both in `d6eb00d`:

**@coderabbitai (input validation):** Fixed. Replaced `max(1, args.decode_steps)` + `assert` with an explicit `parser.error` so non-positive values fail loudly (and don't disappear under `python -O`):
```python
_n_steps = args.decode_steps
if _n_steps < 1:
    parser.error(f"--decode-steps must be >= 1, got {_n_steps}")
```
(The upper-bound `decode_steps < MAX_SEQ` assert is gone entirely — the pool is now enlarged by `decode_steps`, so there's no MAX_SEQ ceiling to guard.)

**@gemini-code-assist (KV sync / H2D copy):** Good catch, clarified in the code. Two points:
- **Timing is unaffected.** The number this tool reports is the device `Effective` span (orch∥sched), which is *device-side* and excludes the host-side `bind.args` H2D copy. The redundant kc/vc H2D shows up in host wall-clock (`simpler_run`), not in `Effective`, and the per-step device read cost correctly tracks the growing `seq_lens`.
- **Correctness scope.** You're right that `kc`/`vc` aren't `InOut`, so device KV writes don't persist across steps — each dispatch re-uploads the prefill KV. That's acceptable for this tool because it measures decode *cost* at growing context (argmax is skipped for N>1); it is **not** a correct autoregressive generation. I removed the "genuine prefill→decode" wording from the docstring/comments and documented this explicitly. Making KV device-resident would need a `decode_fwd` signature change (InOut kc/vc), which is out of scope for this timing harness.

Verified on a2a3 (batch-16, 40L, ctx 3338→3357, `--decode-steps 20`): grows without overflow, 20 STRACE decode invocations, device `Effective` ≈ 37.1 ms median.

PR #796, opened 2026-07-18
perf(qwen3-14b decode): fold RoPE into the CANN FAI attention extern

ChaoWao:
## Summary

- Fold QK-norm and RoPE into the CANN FAI attention extern, removing the
  standalone RoPE dispatch and its producer-to-consumer dependency edge.
- Run the generated RoPE body on its original 32 logical AIV workers; the
  remaining AIV workers skip RoPE but still participate in the mixed-core
  barrier and attention phase.
- Remove the two redundant `dsb(DSB_DDR)` calls around the RoPE-to-attention
  `SyncAll<false>` boundary while retaining the cache-backed metadata barrier.
- Add a contract regression test that ties the C++ worker guard to the Python
  `ROPE_CORES` setting and generated RoPE stride.
- Add a reusable guide for on-device InCore timestamp capture, exact phase
  partitioning, and reconciliation with L2 task timing.
- Record the C220/CANN 9.0.0 no-DSB evidence, validation matrix, near-tie
  diagnosis, and re-evaluation boundary in a dated investigation note.
- Validate the no-DSB candidate in ten synthetic 40-layer-plus-LM-head device
  runs: 9 clean, 1 reference near-tie, 0 hard failures, sampled token 160/160,
  all 24,330,240 logits within `5e-2`, and no NaN or Inf. The synthetic harness
  repeats one generated layer and is not a real-checkpoint accuracy test.

## Related Issues

None.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -25,6 +25,7 @@ Files ending in `_draft.py` are works-in-progress and excluded from CI.
 - `docs/compile-runtime-workflow.md` — what `python <kernel>.py -p <platform>` does end-to-end (compile passes/codegen → input gen → golden → runtime → validate)
 - `docs/debugging.md` — debugging playbook: pypto/ptoas errors, `golden_data` replay, `runtime_dir` reuse, runtime-hang device logs, args-dump / dep-gen
 - `docs/performance-tuning.md` — L2 (inter-kernel) and L1/L0 (intra-kernel) tuning: swimlanes, PMU, buffer-occupancy / perf-hint reports
+- `docs/incore-timestamp-profiling.md` — on-device multi-core phase timestamps for fused extern kernels: per-core capture, barrier diagnostics, and exact L2-reconciled partitions
 - `docs/precision-tuning.md` — keeping a kernel numerically faithful: `pl.cast` rounding modes vs torch, dtype alignment, fp32 intermediates / no double-cast, quant schemes, the `error_distribution` threshold sweep, and real-weight testing
 - `docs/cce-extern-kernel-guide.md` — writing hand-written mixed (cube+vector) CCE kernels behind `pl.jit.extern`: the persistent-kernel runtime model, the tensors-first/scalars-last arg-packing trap, UB/`TPipe`, `SyncAll<false>` cross-core barriers, GM scalar coherency, and the on-device bisection methodology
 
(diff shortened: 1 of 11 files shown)


ChaoWao:
## Investigation snapshot — attention still garbage (BLOCKED)

Pushed `docs/investigations/2026-07-qwen3-rope-fai-fusion-accuracy.md` with the full localization. TL;DR:

**Decisively proven (a2a3, partial dump via `pl.dump_tag` + `enable_dump_args=1`):**
- The RoPE writes **bit-exact-correct** K/V to the paged cache, and `metadata.kv_lengths/cumulative_q` are correct — yet the FA output is garbage. So it's **not** the rope, coherency, metadata, or write location.
- **RoPE no-op isolation** (`if (false) rope_qkv(...)`, keep barrier+FA): the FA becomes **perfect** (`attn_out == seed V` bit-exact). → the fusion *structure* is fine; **the RoPE's *execution* corrupts the FA compute, globally.**

**Ruled out on device (no change):** read-coherency (full `dcci` = no-op), FD-combine (force NonFd), UB/TPipe reset, vector-mask reset, `noinline`, restricting rope to lanes 0-31, and removing `SyncAll<false>`. The compiled FA path uses `resource.ptoTopology` (not the broken native `GetSubBlockIdx`).

**Masking traps** (why it looked fine): fixture `wo` is tiny (out≈residual) + LM-head argmax/logits are robust → single-layer golden false-passes even with attention broken (guide §6). The sharp signal is a strict `ratio_allclose` on `attn_out` (99.7% wrong).

**Sub-result:** hand-written rope replaced with the verbatim pypto-generated `rope_qkv` (rope now bit-exact); the FA blocker is independent of the rope source.

**Next:** in-`rope_qkv` `return;` bisection (guide §7.3) + partial-dump of `attn_out` to bracket the corrupting statement, then reset the shared pto↔AscendC hw/SPR/framework state it touches.

ChaoWao:
### ✅ Root-caused, fixed, and verified on-device (a2a3)

Confirming the fix in `8b123fe` from an independent debugging pass (partial
`dump_tag` on the fused op output vs. what `out_proj` reads):

- The FA output buffer was always **correct**; `out_proj` consumed a *different*
  buffer that was bit-exact **RoPE'd Q** (`q_tnd`). A `pl.jit.extern` returns its
  **first mutable arg**; folding the rope made `query: pl.InOut` lead the
  signature, so `attn_out = paged_attention_rope_cce(...)` bound to `query`, not
  `out`. Placing `out` first (+ `WithRope`-conditional ABI remap) fixes it.
- Ruled out en route (each had *no* effect, so not the cause): a writer-side
  `PipeBarrier<PIPE_ALL>+dsb` drain, dropping `allow_early_resolve`, and `out:
  pl.InOut` — the wrong value was deterministic, i.e. a wrong-buffer bind, not a
  write-back race or early dispatch.

**Strict per-element `out` golden** (`ratio_allclose atol=rtol=3e-3, ≤2%
outliers`, *not* the argmax/logits proxy):

| case | result |
| --- | --- |
| `seq_len=1`, seed 1234 | PASS |
| `seq_len=1`, seed 7 | PASS |
| varied multi-block KV (seq_lens up to 3756), seed 1234 | PASS |

`test_fused_attention_declares_real_output_first` passes.

### ⚠️ Unrelated pre-existing CI note

`test_loaded_kernel_signatures_match_contract_arg_counts` and
`test_compile_arg_builders_follow_loaded_stage_specs` fail on the **prefill**
stage (`prefill_fwd` has 25 params, its contract spec lists 24). Both test
functions pre-date this PR (they exist on `main`); this PR touches neither
`prefill_fwd` nor the contract registry, so the drift is pre-existing (last
prefill change on `main` is #799) and should be triaged separately — it does
not gate this fix.

PR #804, opened 2026-07-20
perf(dsv4 hc_pre): widen linear split-K LINEAR_OK 4->8 (align with hc_head)

wangqin1723-max:
## What
Align `hc_pre_linear`'s split-K factor with `hc_head`: `LINEAR_OK 4 -> 8`. Each cube task's K reduction halves (4096 -> 2048); decode task count doubles (4 -> 8). `K_CHUNK` (256) and L0B pressure are unchanged, and the result is **bit-identical** (same atomic-add reduction, same K order per slice). Matches `hc_head`'s `LINEAR_OK = 8`.

## Why
- **Consistency**: `hc_head` already uses `LINEAR_OK = 8` for the same linear split-K structure; aligning `hc_pre` removes a needless divergence.
- The `hc_pre_linear` kernel itself gets faster (shorter `matmul_acc` chain).

## A/B (DSv4 moe EP2 decode, even primary dies, 6 interleaved rounds)

| metric | OK=4 | OK=8 |
|---|---|---|
| `hc_pre_linear` kernel Exec | 9.75 us | **6.51 us (-33%)** |
| `comb_sinkhorn` Exec | 14.44 us | 14.81 us (unchanged) |
| hc_pre decode wall (isolated) | 45.35 us | 44.15 us (-2.6% mean) |

The linear kernel speedup is real (-33%), but it does **not** reliably move the end-to-end wall: `linear` (6.5us) sits below `rms` (9.8us) and well below the actual pole `comb_sinkhorn` (14.7us, the 20-iter serial Sinkhorn recurrence). The -2.6% wall mean is **not statistically significant** at n=6 (paired diffs are scattered around zero, p ~ 0.19).

Kept for `hc_head` consistency and to pre-stage the split-K widening for when a later change shifts the hc_pre pole onto `linear`.

## Validation
- `python models/deepseek/v4/hc_pre.py -p a2a3 --enable-l2-swimlane` -- golden PASS (decode B=4 S=2 + prefill T=128; `x_mixed` / `post` / `comb` all within tolerance).

### DIFF
--- a/models/deepseek/v4/hc_pre.py
+++ b/models/deepseek/v4/hc_pre.py
@@ -119,7 +119,12 @@ CAST_K_SPMD = 2048  # cast K per spmd block: decode fans the BF16->FP32 cast ove
 # partials, filling idle cubes at small T (decode: 1 token-tile -> LINEAR_OK
 # cube tasks) and shortening each task's matmul_acc chain. Higher OK fills more
 # decode cubes; prefill (8 token-tiles) packs OK*8 tasks into waves of ~24.
-LINEAR_OK = 4
+# A/B 2026-07: raised 4 -> 8 to test whether a smaller/faster linear scope
+# advances the linear -> comb_sinkhorn handoff (comb_sinkhorn reads mixes_raw).
+# Each task's K halves (4096 -> 2048); K_CHUNK (256) and L0B pressure unchanged.
+# Counter-pressure: 2x atomic-add producers on the same mixes_raw[0:16,0:32]
+# tile at the handoff. Matches hc_head's LINEAR_OK = 8.
+LINEAR_OK = 8
 LINEAR_K_PER_SPLIT = HC_DIM // LINEAR_OK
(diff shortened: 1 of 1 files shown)


zhangqi-chen:
Closing as stale — this is an A/B experiment that was never concluded, on a path that no longer exists.

The diff describes itself as such: "A/B 2026-07: raised 4 -> 8 to test whether a smaller/faster linear scope advances the linear -> comb_sinkhorn handoff", and names the counter-pressure (2x atomic-add producers on the same `mixes_raw` tile) without resolving it. No measurement was posted.

State on main `e058d747`: both `models/deepseek_v4_flash_mtp/hc_pre.py:55` and `models/deepseek_v4_flash_dspark/hc_pre.py:40` still have `LINEAR_OK = 4`, so the widening was not adopted anywhere. The target file `models/deepseek/v4/hc_pre.py` was removed by #886 (`models/` flattened to one directory per model build, 2026-08-04).

If the split-K width is worth revisiting, a fresh PR with the wall-time A/B on the current `hc_pre.py` would settle it in one round.

PR #809, opened 2026-07-21
perf(dsv4/moe): predicate empty routed-expert stages

wangqin1723-max:
## Summary
- Predicate each routed-expert SPMD stage (gate / up / act / quant / w2) on `recv_expert_count[local, 0] > 0` so experts with zero tokens skip all compute.
- Carry per-expert TaskIDs (`gate_tids` / `up_tids` / `act_tid` / `quant_tid`) across the split compute passes and wire them as explicit `deps=[count_tid, gather_tid, ...]` edges, so predicated tasks chain correctly while auto dependency tracking still covers tensor reads/writes.
- `dispatch()` returns its `_meta_tid` / `_gather_tid` and threads them into `expert_routed` as the count/gather producers.
- Flatten `recv_x` to a 2D ND view (`recv_x_flat`) to stop PTOAS inferring NZ on the 3-D view when `RECV_MAX == 32` (EP=4), which conflicted with the row-major matmul input layout.
- Switch `exp_h_q` from `pl.at(CORE_GROUP)` to `pl.spmd(1)` for a uniform predicated scope; lower `n_routed_experts` 256 -> 128 (16/rank) so more experts hit the empty path.

This branch also carries the routed-expert stage-hoisting from #738 (launch each stage once per expert; read the routed row count inside each stage), on top of which the predicate work is layered.

## Related Issues
Stacks on #738.


zhangqi-chen:
Closing as stale — the proposal was not adopted and the target paths are gone.

Empty routed-expert stage predication is not present on main `e058d747`, and the four files this PR touches (`models/deepseek/v4/{config,expert_routed,gate,moe}.py`) were removed by #886, which flattened `models/` to one directory per model build on 2026-08-04. The PR has been `CONFLICTING` since, with no update for over six weeks.

The MoE path has also moved substantially since July — the routed-expert kernel was restructured across #1039-era work — so the diff would need rewriting rather than rebasing.

Reopen as a fresh PR against `models/deepseek_v4_flash_mtp/expert_routed.py` if empty-stage predication is still worth the dispatch savings; a decode wall-time number on the skewed routing distributions would make the case quickly.

PR #811, opened 2026-07-21
perf(dsv4/moe): tune early task resolution

Little-oil:
## Summary

- tune early task resolution across DeepSeek V4 HC pre-processing, dispatch/combine, and routed/shared expert stages
- add dependency-only dummy tasks for x_norm_quant and dispatch_meta scheduling experiments
- move early-resolution markers between shared/routed W2 stages to improve overlap opportunities

## Notes

- dispatch_push still does not early-dispatch after adding the dispatch_meta seed dummy; the source records this as a follow-up scheduling point

## Validation

- Python syntax compilation for all five modified model files
- Ruff checks for all five modified model files
- git diff whitespace check
- python models/deepseek/v4/moe.py --compile-only --ep 2 -p a2a3 -d 0,1


Little-oil:
moe 500 → 484 
测试不稳定！最好的100次L1平均值是479

PR #813, opened 2026-07-22
Fix hc_pre reduction scratch tile shapes

lyfne123:
## Summary\n- Size hc_pre row-reduction scratch tiles to the padded physical input shape.\n- Keep the syncall and separate implementations aligned with PyPTO validation.\n\n## Related Issues\n- Companion to hw-native-sys/pypto#2082 and hw-native-sys/pypto#2099


lyfne123:
Hardware validation against the companion PyPTO change passed on A2/A3:

- Entry: models/deepseek/v4/decode_attention_swa.py -p a2a3
- Task: task_20260721_185110_268194629247
- Result: kv_cache PASS, x_out PASS, overall PASS

PR #821, opened 2026-07-23
CI: disable all jobs, add A5 runner bring-up smoke job

luohuan19:
## Summary

A new A5 server has been added as a CI runner; this gates the workflow down to a single job that only proves the runner can be picked up.

- Disable every existing job in `ci.yml` (`pre-commit`, `unit-tests`, `detect-changes`, `sim`, `a2a3`, `serving-qwen`, `serving-deepseek`) with `if: false`, kept as `false && <original expression>` where a condition already existed, so the original gating logic survives verbatim and re-enabling is a prefix removal.
- Add `a5-runner-smoke`, structurally a copy of `daily_ci.yml`'s `model-tests-a5`: same `[self-hosted, linux, arm64, npu-a5]` labels, 120 min timeout, `checkout` working directory and env block, and the same step sequence (sparse-fetch action defs, `setup-ci-job` with `needs-device: 'true'`, `kill-orphaned-tasks`, `build_output` cleanup, results upload).
- Replace that job's `Run V4-Pro tests (a5, via task-submit)` body with `echo "Do something"`, keeping the step name and its `PYPTO_*` / `PTO2_RING_*` env.
- Use a dedicated `a5-smoke` build namespace so it does not share the daily job's `a5-daily` artifacts, and set `if-no-files-found: ignore` on the upload since the smoke step emits no `results.tsv`.

### DIFF
--- a/.github/actions/setup-ci-job/action.yml
+++ b/.github/actions/setup-ci-job/action.yml
@@ -298,7 +298,7 @@ runs:
     - name: Check NPU
       if: ${{ inputs.needs-device == 'true' }}
       shell: bash
-      run: npu-smi info
+      run: task-submit --run "npu-smi info"
 
     # Namespace the cache by pto-isa commit so a pto-isa bump can never serve
     # stale objects compiled against the old ISA headers (see issue #1139):
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,13 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # NOTE: every pre-existing job below is temporarily disabled (`if: false`,
+  # kept as `false && <original expr>` where a condition already existed) while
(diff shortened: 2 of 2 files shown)


zhangqi-chen:
Closing — the A5 bring-up this PR was staging for is done, by other means.

`daily_ci.yml` on main `e058d747` now carries real A5 coverage on `[self-hosted, linux, arm64, npu-a5]` runners: a `model-tests-a5` device job running the V4-Pro tests via task-submit, and a `DeepSeek-V4-Flash EP8 end-to-end (a5)` job running the 8-card token loop. #1096 and #1147 have since tuned which jobs follow HEAD and routed a5sim to examples.

The "disable all jobs" half was a temporary bring-up measure and must not land now that those jobs are the coverage. The PR has also been `CONFLICTING` against the current workflow files for six weeks.

Thanks for the bring-up work — closing because it arrived rather than because it was rejected.

PR #827, opened 2026-07-24
fix(deepseek-v4-flash): fit decode buffers for PTOAS planning

Little-oil:
## Summary

- replace a redundant zero-fill/column-expand/add sequence with `col_expand_add`
- keep empty sparse-attention block zeroing to a one-row Vec tile to avoid reserve-buffer fragmentation

The branch is rebased onto the DeepSeek V4 split; these changes now target
`models/deepseek/v4-flash`. The row-reduction scratch sizing from the original
branch is already present upstream via #813.

## Dependency

Requires [hw-native-sys/pypto#2128](https://github.com/hw-native-sys/pypto/pull/2128).

## Testing

- DeepSeek V4 Flash decode with PTOAS planning and PTOAS 0.48 on devices 4/6:
  PASS in 172.17s (`task_20260723_190106_12367731265`)
- DeepSeek V4 Flash decode with PyPTO planning and PTOAS 0.48 on devices 4/6:
  PASS in 172.00s (`task_20260723_190705_14377107555`)
- focused compiler/device suites in the companion PyPTO PR: 234 unit tests and 18 numerical device tests passed
- pypto-lib pre-commit hooks for the rebased diff

The full DeepSeek driver has no `golden_fn` or `golden_data`; full-model runs
cover compilation and dual-device execution, while the focused device tests
perform numerical golden comparison.

### DIFF
--- a/models/deepseek/v4-flash/decode_sparse_attn.py
+++ b/models/deepseek/v4-flash/decode_sparse_attn.py
@@ -364,14 +364,17 @@ def sparse_attn(
                         sparse_blk_li[qk_row : qk_row + H_TILE, 0 : 1] = qk_li[qk_r0 : qk_r0 + H_TILE, 0 : 1]
                         sparse_blk_oi[qk_row : qk_row + H_TILE, 0 : HEAD_DIM] = qk_oi[qk_r0 : qk_r0 + H_TILE, 0 : HEAD_DIM]
             else:
-                qk_oi_zero = pl.full([H_TILE, HEAD_DIM], dtype=pl.FP32, value=0.0)
+                # Keep the empty-block zero tile to one row. A full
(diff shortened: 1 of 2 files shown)


zhangqi-chen:
Closing as stale — not adopted, and the target paths no longer exist.

The buffer fit for PTOAS planning in this PR is not on main `e058d747`, and `models/deepseek/v4-flash/decode_sparse_attn.py` / `decode_sparse_attn_hca.py` were removed by #886 (`models/` flattened to one directory per model build, 2026-08-04). The PR has been `CONFLICTING` since, untouched for over six weeks.

PTOAS has also moved from 0.57 to 0.60 in that window, so the planning constraint this worked around should be re-measured before the same fix is re-applied.

Please open a fresh PR against `models/deepseek_v4_flash_mtp/` if the decode buffers still trip PTOAS planning on the current pin.

PR #846, opened 2026-07-27
Perf: defer DSv4-flash attention RMSNorm reduce past the QKV projection

Hzfengsy:
## Summary

> Stacked on #845. Until that merges, the diff here also shows its commit; review only `Perf: defer DSv4-flash attention RMSNorm reduce past the QKV projection`.

`RMSNorm(x) = (x * gamma) * inv_rms`, and `inv_rms` is a per-token positive scalar. Consumers that are linear in `x`, or invariant to a positive per-token scale (a second RMSNorm, symmetric per-token quant, a top-k), can take the elementwise half directly and fold `inv_rms` in later — so the reduce no longer gates the projection it feeds.

- Split `rms_norm` into `rms_gamma` (elementwise) and `rms_recip` (reduce). `rms_gamma` fans over `D` as well as `T`, which the fused form cannot: there is no cross-D dependency once the denominator is removed.
- CSA / HCA: run both halves, and pass `x_inv_rms` to the compressor and indexer. `qkv_proj_rope` needs no correction — its q_a / kv_a LoRA outputs are re-normalized and its qr quant is symmetric per token, both invariant to a positive per-token scale — so `qr_proj_matmul`, the critical path, no longer waits on the reduce.
- SWA: `qkv_proj_rope` is the only consumer of the norm, so `inv_rms` is never observable and the reduce is dropped entirely rather than deferred.
- `compressor_ratio4` / `compressor_ratio128` / `indexer_compressor`: apply the row scale in `scatter_softmax_pool`, ahead of the `ape` add and ahead of the pool that mixes tokens carrying different `inv_rms`. Kept in that vector-only scope rather than the `kv_score_proj` epilogue, where a vector op would turn a pure-cube task mixed and add a cube->UB->GM round-trip.
- `indexer`: fold the scale into the existing split-K reduce in `weights_proj` rather than adding a pass over `x`.

`rms_norm` is unchanged, so the prefill, MTP and final-norm callers keep the fused form.

Latency from norm-scope start to `qr_proj_matmul` start, per layer (decode, B=4 S=2, a2a3): CSA 51.4 -> 46.3 us, HCA 20.0 -> 14.8 us, SWA 19.8 -> 12.0 us. SWA standalone wall 282.1 -> 271.7 us.

### DIFF
--- a/models/deepseek/v4-flash/decode_attention_csa.py
+++ b/models/deepseek/v4-flash/decode_attention_csa.py
@@ -51,6 +51,7 @@ from hc_pre import hc_pre
 from decode_indexer import indexer
 from qkv_proj_rope import qkv_proj_rope
 from rmsnorm import rms_norm
+from rope_interleave import rope_interleave
 from decode_sparse_attn import sparse_attn
 
 # model config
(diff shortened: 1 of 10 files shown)


zhangqi-chen:
Closing as stale — not adopted, and the ten target files no longer exist.

Deferring the attention RMSNorm reduce past the QKV projection is not on main `e058d747`. All ten files this PR touches live under `models/deepseek/v4-flash/`, which #886 removed on 2026-08-04 when `models/` was flattened to one directory per model build; the PR has been `CONFLICTING` since, with no update for six weeks.

Worth recording for whoever picks this up: the deferred pre-QKV attention norm was subsequently found to help **SWA decode only**, not the CSA/HCA paths — so a re-land should be scoped to that case rather than applied across all three attention paths as this diff does.

Please open a fresh PR against `models/deepseek_v4_flash_mtp/` with that scoping and a wall-time number.

PR #852, opened 2026-07-28
fix(deepseek): return explicit decode output buffers

hashiqiqixian:
### Summary

Return the existing `pl.Out` decode compressor buffers directly from the three DeepSeek V4 Flash inline kernels. This keeps the runtime tensor mapping explicit for orchestration codegen while preserving the existing flattened-view writes.

### Motivation

When the kernels end with `kv = pl.reshape(kv_flat, ...)`, their returned SSA value is the reshape result rather than the declared `pl.Out` parameter. In composed serving compilation, orchestration codegen cannot map that return to one of the callee's Out/InOut parameters and rejects the IR with `IRProperty::ReturnParamsExplicit`.

### Proposed Change

- Remove the trailing reshape reassignment in the ratio-4, ratio-128, and indexer decode compressors.
- Continue writing through `kv_flat`, which is a view of the same caller-owned `kv` buffer.
- Return the declared `kv` output parameter directly.

### Boundaries

This PR does:
- Preserve explicit output-buffer identity for DeepSeek V4 decode orchestration.

This PR does not:
- Change kernel math, tensor shapes, cache layout, or serving scheduling.
- Add serving CLI, request state, target verification, or acceptance accounting; those are owned by hw-native-sys/pypto-serving#120.

### A/B validation

The same 8-NPU DeepSeek V4 Flash W8A8 K=9 HTTP test was run on devices 8-15 with only the pypto-lib revision changed:

- Control: pypto-serving `cb77f085affea7104398af9409643d0d57fee6bd` + pypto-lib `186e1d85911f37ead07b8e74fd12734f23c947f0` -> task `task_20260728_210239_293720714497`, exit 1. `decode_fwd` failed with `cannot map return of callee 'rmsnorm_rope_cache_write' ... IRProperty::ReturnParamsExplicit`.
- Fix: pypto-serving `9fc078be026050d89b0d7428ba661a42e738ad50` + pypto-lib `864b05396ec72e2bb42f5f77032a17f83e1794f8` -> task `task_20260728_222400_233761715500`, exit 0. Server became healthy and K=9 returned the expected 10-token text.
- Focused CPU coverage on the fixed revision: `98 passed, 4 warnings`.

The fixed NPU run also compiled `decode_fwd`, `mtp_prefill_fwd`, and `mtp_decode_layer`, loaded all 43 layers, and released devices 8-15 after completion.

### DIFF
--- a/models/deepseek/v4-flash/decode_compressor_ratio128.py
+++ b/models/deepseek/v4-flash/decode_compressor_ratio128.py
@@ -290,7 +290,6 @@ def compressor_ratio128(
                     kv_flat[global_c_idx * s_dim : global_c_idx * s_dim + 1, :] = kv_row
                     cmp_kv_cache_flat[cmp_row : cmp_row + 1, :] = pl.cast(kv_row, target_type=pl.BF16, mode="rint")
 
-    kv = pl.reshape(kv_flat, [b_dim, s_dim, HEAD_DIM])
     return kv
 
 
--- a/models/deepseek/v4-flash/decode_compressor_ratio4.py
+++ b/models/deepseek/v4-flash/decode_compressor_ratio4.py
@@ -261,7 +261,6 @@ def compressor_ratio4(
                     kv_flat[c_idx * S : c_idx * S + 1, :] = kv_row_fp32
                     cmp_kv_cache_flat[cache_row : cache_row + 1, :] = pl.cast(kv_row_fp32, target_type=pl.BF16, mode="rint")
 
-    kv = pl.reshape(kv_flat, [B, S, HEAD_DIM])
     return kv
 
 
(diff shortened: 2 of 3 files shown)


hashiqiqixian:
Closing this PR because we do not yet have an A/B reproduction showing that the output-return cleanup is required by pypto-serving arbitrary-depth MTP. The simulator stall reproduced on both revisions and does not establish that dependency. The serving PR will be re-pinned to the upstream parent revision; this cleanup can be reconsidered independently with a focused reproducer.

hashiqiqixian:
Current CI status for `864b053` (run https://github.com/hw-native-sys/pypto-lib/actions/runs/30426803327):

- PASS: pre-commit, unit-tests, a2a3, serving-deepseek, CodeRabbit.
- SKIP: serving-qwen (change detection).
- STUCK: both simulator jobs enter the existing PTO2 scheduler `S1:running-stalled` signature while running the DeepSeek decode-layer path, then are cancelled by the ~30-minute job timeout. a2a3sim reports `completed=88/113`; a5sim reports `completed=88/112`.

This is the same simulator-stall class previously reproduced across attempted revisions, while the exact 8-NPU serving A/B now establishes the three-line output-identity fix: control `186e1d8` fails `ReturnParamsExplicit`; fix `864b053` serves K=9 successfully. No simulator-specific code change is being added to this focused PR.

PR #856, opened 2026-07-29
Fix: tile prefill_indexer qr_proj over rows to avoid a bad INT32 matmul

lwDavid:
The Q projection accumulated all T=128 token rows in one
[T, Q_OUT_TILE] INT8 x INT8 -> INT32 matmul. On a5 that form silently
produces wrong products: the dequantised qr_proj comes out correlated
0.086 with the reference, which corrupts `score` at every visible
position (2010 of 2016) while leaving per-row norms intact, so it
surfaces as a precision-looking failure rather than a crash.

Accumulate one QR_PROJ_ROW_TILE (16) row block at a time instead, which
is the shape decode_indexer already uses -- decode never hit this because
its T fits in a single 16-row tile.

Localised by exporting the kernel's intermediates and diffing them
against a host recomputation: the weights projection matches the
reference exactly (correlation 1.00000) while qr_proj, the first stage,
is already wrong, and its NOPE half -- which RoPE never touches -- is
wrong on its own. Ruled out along the way: RoPE convention and its
cos/sin token mapping, Hadamard orientation, quantisation granularity,
and the weights path.

This is a workaround for a defect below pypto-lib, not a fix for it. The
emitted tiles all fit a5's on-chip limits (L0A 16 KB, L0B 32 KB, L0C
128 KB against 64/64/256 KB), so capacity is not the cause.

prefill_indexer now passes 4/4 on a5; it previously failed every run.

### DIFF
--- a/models/deepseek/v4-pro/prefill_indexer.py
+++ b/models/deepseek/v4-pro/prefill_indexer.py
@@ -137,18 +137,29 @@ def prefill_indexer(
     qr_proj = pl.create_tensor([T, IDX_N_HEADS * IDX_HEAD_DIM], dtype=pl.FP32)
     for idx in pl.spmd(IDX_N_HEADS * IDX_HEAD_DIM // Q_OUT_TILE, name_hint="prefill_idx_qr_proj"):
         o0 = idx * Q_OUT_TILE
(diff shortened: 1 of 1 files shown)


lwDavid (commented):
Reviewed. This is a temporary plan to avoid bugs(https://github.com/hw-native-sys/pto-isa/issues/214).

PR #866, opened 2026-07-30
Fix: align gate golden quantization scale with device

yanghaoran29:
- Use an FP32 tensor numerator for the gate golden quantization scale
  so Torch follows tensor-to-tensor division like device `pl.div`.
- Avoid reciprocal-multiply rounding differences at INT8 half-integer
  boundaries without changing the device kernel.
- Keep the generated gate outputs unchanged while making the golden
  reference match the device FP32 rounding semantics.

### DIFF
--- a/models/deepseek/v4-pro/gate.py
+++ b/models/deepseek/v4-pro/gate.py
@@ -341,7 +341,7 @@ def _per_token_int8_quant(x_bf16):
     import torch
     x_f32 = x_bf16.float()
     amax = x_f32.abs().amax(dim=-1, keepdim=True).clamp_min(INT8_AMAX_EPS)
-    scale_q = INT8_SCALE_MAX / amax
+    scale_q = torch.full_like(amax, INT8_SCALE_MAX) / amax
     scaled = x_f32 * scale_q
     x_i8 = torch.round(scaled).to(torch.int32).to(torch.float16).to(torch.int8)
     scale_dq = (1.0 / scale_q).reshape(-1)  # [T]


yanghaoran29:
- Fixed the Torch golden INT8 quantization scale calculation for DeepSeek V4 Pro gate to align its semantics with the FP32 computation of `pl.div` on the device side.

- Original implementation divides a Python scalar by a Tensor:
  ```python
  scale_q = INT8_SCALE_MAX / amax
  ```
  In the current Torch pipeline, this expression is equivalent to computing the reciprocal first then multiplying by 127, which introduces a difference of 1~2 FP32 ULPs compared to `pl.div` on the device.

- Explicitly construct the numerator as an FP32 Tensor matching the dtype and shape of `amax`:
  ```python
  scale_q = torch.full_like(amax, INT8_SCALE_MAX) / amax
  ```
  This forces Torch to perform tensor/tensor FP32 division. Across four captured groups of related tokens, the revised scale values match device outputs bit-for-bit.

- Quantization boundary analysis shows `amax` values are identical between device and Torch, yet the original golden implementation exhibits minor FP32 least-significant-bit discrepancies in scale values. Examples:
  ```
  amax = 4.59375
  device scale = 27.6462593079
  Torch scale  = 27.6462574005

  amax = 3.484375
  device scale = 36.4484291077
  Torch scale  = 36.4484329224
  ```

- The original golden implementation produced 39 INT8 mismatches out of 114688 points on a fixed failure case. 38 mismatches stem solely from scale differences, with only 1 mismatch originating from upstream BF16 variance in `hc_pre`.

- Typical quantization boundary case:
  ```
  device scaled = 63.5000000 -> INT8 64
  Torch scaled  = 63.4999962 -> INT8 63
  ```
  The absolute gap is approximately 3.8e-6, yet the values fall on opposite sides of the half-integer boundary for `rint`, resulting in a 1-unit INT8 discrepancy in the final quantized value.

- After TOPK=6 routing, the 39 INT8 mismatches in gate expand strictly to 234 mismatches in `recv_x`. These errors further propagate through dense W1/W3 weights, activations, and W2 computations of routed experts to contaminate a large number of final output channels.

- Regenerating golden outputs with the same fixed MoE input reduces the out-of-threshold ratio of `x_next` for full EP2 MoE from:
  ```
  27336 / 458752 = 5.9588% (FAIL)
  ```
  to:
  ```
  9800 / 458752 = 2.1362% (PASS)
  ```
  Mainline criteria: single-point relative error threshold = 0.003, with a maximum allowed out-of-threshold ratio of 5%.

- Re-run all affected test cases post-fix:
  - DeepSeek V4 Pro gate: PASS
    - x_norm_i8: PASS
    - x_norm_scale: PASS
    - indices: PASS
    - weights: PASS
    - x_next: PASS

  Gate validation task ID:
  `task_20260730_202139_20242049386`

  MoE EP2 validation task ID:
  `task_20260730_202139_202421218783`

- This change only modifies the Torch golden reference implementation. No adjustments are made to device kernels, quantization formulas, model outputs, or runtime behavior.

lwDavid (approved):
Reviewed.

PR #878, opened 2026-07-31
Refactor: express the decode matmuls' live rows with valid_shape

lyfne123:
## Summary

- `decode_fwd` pads its pipeline to `BATCH_PAD` rows and feeds every matmul the full 16-row A operand, relying on the padding rows being *harmless* rather than on them being *excluded*. That invariant holds only because every op between the input `fillpad` and the LM-head store is row-independent; any future cross-row reduction would break it silently. Wrap each matmul's A operand in `set_validshape(..., batch, K)` so "only `batch` rows are live" is stated where the rows are consumed. This also unblocks a partial tail chunk, which is how a batch above `BATCH_PAD` would have to be served.
- Wrap the 20 A operands across q/k/v_proj, out_proj's two waves, gate/up's critical and deferred waves, and down_proj. The accumulators are left full-width: they are zero-seeded, and trimming the store would replace a determinate `0` in the padding rows with an indeterminate value. `decode_layer_a8w8` already wrapped its operands this way, but with `ACTIVE_BATCH == BATCH_PAD`, so that path had never run below full width.
- Switch off pypto's test-only `compile_for_test()`: pypto#2230 replaced it with the public `JITFunction.lower()`, which runs the pass pipeline without codegen, caching, or artifacts — what these three compile-only paths already wanted. CI builds pypto from an unpinned shallow clone of `main`, so this is not optional: `decode_fwd`'s `--smoke` path is what the `sim (a2a3sim)` / `sim (a5sim)` jobs run.

**Requires pypto >= `6234d0e` (pypto#2231).** Before that fix, `set_validshape` did not inherit its input's memory space, so the Mat demand from the matmul stopped at the wrapper, the A operand was materialised in Vec, and the resulting Vec->Mat `tile.move` flipped each scope to MIXED. `ExpandMixedKernel` then split every matmul into an AIC/AIV pair: 994 -> 2494 scheduled tasks and +89% device work. Filed as pypto#2227.

## Verification

Validated on a2a3 with pypto `6234d0e`:

| Check | Result |
| --- | --- |
| single-layer golden (batch 16) | PASS |
| `--validate-fwd -b 1` | 1/1 |
| `--validate-fwd -b 8` | 8/8 |
| `--validate-fwd -b 16` | 15/16 — the pre-existing row-11 near-tie (sample 16/16, logits 100% within 5e-2) |
| `decode_layer_a8w8` smoke | PASS |
| compiled program | 57 functions, **zero** matmul-side `_aiv` halves — identical to the pre-valid_shape baseline |

Cost is unchanged: at batch 1 over n=3 paired runs with forced fresh compiles, makespan -0.60% and summed task time -0.84%, both well inside the run-to-run spread (padded spread 162 us vs a 18 us delta). Nothing was expected to get faster — the cube fractal minimum is M=16, so the padding rows were already free; the fix removed a cost that should never have existed rather than creating a saving.

### DIFF
--- a/models/qwen3/14b/decode_fwd.py
+++ b/models/qwen3/14b/decode_fwd.py
@@ -368,7 +368,7 @@ def _decode_layer(  # noqa: PLR0913 — model signature is intrinsic
             for n_sub in pl.range(N_SUB):
                 n0 = q_n_region + n_sub * TN
                 q_acc = pl.matmul(
-                    normed_in[:, q_k_base : q_k_base + TK],
+                    pl.tensor.set_validshape(normed_in[:, q_k_base : q_k_base + TK], batch, TK),
                     wq[layer_hidden_base + q_k_base : layer_hidden_base + q_k_base + TK, n0 : n0 + TN],
                     out_dtype=pl.FP32,
                 )
@@ -376,7 +376,7 @@ def _decode_layer(  # noqa: PLR0913 — model signature is intrinsic
                     kk = q_k_base + kc * TK
                     q_acc = pl.matmul_acc(
                         q_acc,
-                        normed_in[:, kk : kk + TK],
+                        pl.tensor.set_validshape(normed_in[:, kk : kk + TK], batch, TK),
                         wq[layer_hidden_base + kk : layer_hidden_base + kk + TK, n0 : n0 + TN],
                     )
                 q_proj = pl.assemble(q_proj, q_acc, [0, n0], atomic=pl.AtomicType.Add)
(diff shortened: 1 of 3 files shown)


lyfne123:
Superseded by a narrower PR: the valid_shape half is dropped.

Measurement showed it was not worth it — no speedup (the cube fractal minimum is M=16, so the padded rows were already free), a ~2.34% makespan cost at batch 16 (n=8, both pair orders), and, as noted in review, the part that actually scales with batch is the FAI attention, which lives in the CCE extern and is untouched by a pypto-level valid_shape. It also pulled in a hard pypto >= 6234d0e floor.

Keeping only the compile_for_test() -> lower() fix, which is independent and unblocks CI.

PR #879, opened 2026-07-31
Fix(deepseek-v4 pro): let MoE post_ffn use automatic dependency tracking

yanghaoran29:
## Summary
- Remove `manual_dep=True` from MoE `post_ffn` so the framework tracks the real `hc_pre` → `hc_post` producer/consumer edge.
- Keeps `post_ffn` from being treated as an independent buffer that can race under EP scheduling.

## Test plan
- [x] `moe.py -p a5` with fixed `moe_scale_aligned_fixture`, 5 runs across device pairs `0,1` / `2,3` / `1,2`
- [x] All 5 PASS under mainline gate (`rdiff>0.3%`, allow ≤5%); over-threshold ratio ≈ 2.13%–2.14%
- [ ] CI / review confirmation on a5 EP2 MoE

Made with [Cursor](https://cursor.com)

### DIFF
--- a/models/deepseek/v4-pro/moe.py
+++ b/models/deepseek/v4-pro/moe.py
@@ -438,7 +438,7 @@ def moe(
 ) -> pl.Tensor[[T, HC_MULT, D], pl.FP32]:
     # Non-output intermediates allocate locally, in their producer's scope.
     x_mixed = pl.create_tensor([T, D], dtype=pl.BF16)
-    post_ffn = pl.create_tensor([T, HC_MULT], dtype=pl.FP32, manual_dep=True)
+    post_ffn = pl.create_tensor([T, HC_MULT], dtype=pl.FP32)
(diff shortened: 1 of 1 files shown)


zhangqi-chen:
Closing as stale — not adopted, and the target path is gone.

This one-line change drops `manual_dep=True` from `post_ffn` so MoE picks up automatic dependency tracking. On main `e058d747`, `models/deepseek_v4_pro/moe.py:483` still reads `post_ffn = pl.create_tensor([T, HC_MULT], dtype=pl.FP32, manual_dep=True)` — the change never landed. The file it targets, `models/deepseek/v4-pro/moe.py`, was removed by #886 on 2026-08-04.

The change is one line, so re-opening it against the current path costs nothing. Worth flagging for whoever does: a `manual_dep` tensor gets no automatic dependency edge at all, so dropping the flag changes which edges the compiler synthesizes around `post_ffn` — the re-land should come with a device run of the v4-pro MoE path, not just a compile.

Issue #858, opened 2026-07-30
[Bug] a5: two in-budget vector tasks fault with 'VEC access UB out of bounds' when co-resident (prefill_sparse_attn, surfaces as S1:running-stalled)


lwDavid:
### Diagnosis

**Below pypto-lib** — most likely pypto codegen or the simpler runtime; not the
kernel. Two `pl.spmd` vector tasks that are individually within a5's UB budget
take an AICore "VEC access UB out of bounds" fault when they go resident on-core
together. The compiler's on-chip allocation is byte-identical between a faulting
build and a passing one, and the fault flips on pure timing, so nothing in the
IR distinguishes the two cases.

Filed here rather than against pto-isa because I cannot yet attribute it to a
specific layer, and I do not want to send anyone chasing the wrong repo.

### Description

`models/deepseek/v4-pro/prefill_sparse_attn.py` faulted on a5 in **16 of 18
runs** at the merged-main revision. #857 works around it; this issue is for the
underlying defect.

**Reproduce** (at `1f97429`, i.e. before #857):

```bash
task-submit --run --device 0 --timeout 800 --max-time 0 \
  "source <env>.sh && export PYTHONPATH=<repo>:$PYTHONPATH && \
   cd <repo>/models/deepseek/v4-pro && \
   python prefill_sparse_attn.py -p a5 -d \$TASK_DEVICE"
```

Expect a fault in roughly 9 of 10 runs.

#### It is not a hang, despite what the classifier says

The case had been recorded as a device hang on the strength of:

```
sched_error_code=100 SCHEDULER_TIMEOUT
sub_class=S1:running-stalled (detail=1) completed=3/312 running=2 ready=0
  waiting=307 orch_done=1 stuck_task_id=4294967299 stuck_core=2
```

Those counters are a *consequence*. Timestamps from one run:

```
20:06:30.583  last setup line before kernel launch
20:06:30.639  EZ9999 AICore exception            <-- 56 ms after launch
20:07:08.716  SCHEDULER_TIMEOUT                  <-- the 38 s timer expiring
```

Every faulting run carries a second errcode that names the real fault; **no**
passing run does:

```
The extend info: errcode:(341) errorStr: The address for VEC to access UB is
out of bounds. subErrType: 0x4.
[FUNC:ProcessDavidStarsCoreErrorInfo][FILE:device_error_proc_c.cc][LINE:750]
```

`errcode:(0) timeout or trap error` on its own is the drain artifact. Likewise
[...]

PR #891, opened 2026-08-03
Fix: migrate pl.load valid_shapes to valid_shape

lwDavid:
## Summary

pypto [`8a429413`](https://github.com/hw-native-sys/pypto/commit/8a429413) — *"refactor(dsl): Unify load valid_shape naming"* (pypto#2233) — renamed the `pl.load` keyword `valid_shapes` → `valid_shape` and removed the old spelling with, in its own words:

> The removed `valid_shapes` keyword **intentionally has no alias or deprecation path**.

So every `pl.load` that passes it now fails at compile time:

```text
Error: pl operation 'load': load() got an unexpected keyword argument 'valid_shapes'
  --> models/deepseek/v4-pro/hc_pre.py:519:8
```

This migrates the only two files in the repo that use it — 17 sites each in
`models/deepseek/v4-pro/hc_pre.py` and `models/deepseek/v4-flash/hc_pre.py`.

## Why it matters more than 34 lines suggests

`hc_pre` is inlined by most of the attention path, so this one keyword took down
**7 of the 27 V4-Pro single-card cases** at compile time — `hc_pre` itself plus
`decode_attention_{csa,hca,swa}` and `prefill_attention_{csa,hca,swa}`.

## Scope: keyword rename only

- Upstream states the fourth positional `tile.load` operand and the serialized IR
  operand order are unchanged, so this is not a semantic change.
- Masking the keyword makes every `+`/`-` line in this diff byte-identical —
  nothing else moved.
- The singular `valid_shape=` already present in both files belongs to `pl.slice`,
  which has always used the singular spelling. Those 5 sites per file are
  deliberately untouched.
- `grep -rn valid_shapes --include=*.py` over the repo is empty after this change.

## Verification

Real A5 hardware, serial (one JIT build in the tree at a time), on a tree pinned to
what pypto main pins:

| Component | Version |
|---|---|
| pypto | `385dbed3` (main) |
| simpler | `dccb8379` (pypto main's runtime gitlink) |
| pto-isa | `83d01313` (`simpler:pto_isa.pin`) |
| ptoas | v0.48 — see the compatibility note below |
| CANN / driver | 9.1.0-beta.3 / 25.7.rc1.6 |

- The 7 previously compile-broken V4-Pro cases now build and run.
- V4-Pro single-card suite: **26 / 27**. The one exception is
  `prefill_attention_csa`, the pre-existing [#858](https://github.com/hw-native-sys/pypto-lib/issues/858)
  fault (10/20 here, 7/16 historically — unchanged and unrelated to this PR).
- V4-Flash was not run on device; its change is textually identical to V4-Pro's.

## Compatibility — please read before merging

**This PR requires pypto ≥ `8a429413` and hard-breaks any older pypto.** There is
no alias upstream, so the two spellings cannot coexist. `hc_pre.py` will raise the
error above on any pypto that predates pypto#2233.

CI takes pypto from the self-hosted runners' shared conda venv rather than from a
ref pinned in `.github/workflows/`, so **whether CI goes green depends on when that
venv was last refreshed**. If it is still on a pre-#2233 pypto, this PR should wait
for the venv bump rather than be merged and reverted.

## Unrelated but worth flagging

ptoas is held at **v0.48** above, not the **v0.54** that `pypto@385dbed3` pins,
because v0.54 independently breaks this workload: measured in the same tree
changing only `PTOAS_ROOT`, v0.54 gives 12/27 and v0.48 gives 26/27, with all 15
failures recovering. Filed as [PTOAS#1111](https://github.com/hw-native-sys/PTOAS/issues/1111).
That is a separate problem and does not affect this PR — the rename is required on
either assembler.


lwDavid (commented):
Reviewed. API fix.

PR #892, opened 2026-08-04
Add: serve any decode_fwd public batch by chunking into row windows

lyfne123:
decode_fwd was limited to a public batch of at most BATCH_PAD (16): the
attention bridge derived its batch from the seq_lens descriptor, so one
call always covered the whole batch and the FAI tiler rejected anything
past kMaxBatch. Keep the pipeline padded at 16 rows and run a larger
batch as ceil(batch / BATCH_PAD) consecutive row windows instead.

- Give the CCE bridge an explicit row window. The tiling extern takes
  batch_offset / batch_count and emits a 0-based tiling for that slice;
  the fused rope+attention extern takes batch_offset (args[18]) and
  offsets seq_lens, slot_mapping and the block table by it. The RoPE
  phase now reads its batch back from the emitted tiling, so it and the
  attention phase still cannot disagree about how many rows are live.
- Loop decode_fwd over the windows: each runs the full layer stack and
  the LM head for its rows, while the token embedding and the sampling
  stay outside (both already walk the whole public batch). Windows are
  serialized on the shared metadata/workspace pair by carrying the
  previous window's final writer into the next tiling rebuild.
- rms_lm_head_fp32 takes row_offset / valid_rows instead of seq_lens.
- Drop the decode batch ceiling in the contract; prefill and the
  standalone greedy_sample stage stay capped at the padded width, and
  limits["batch"] now documents one decode window rather than a maximum.
- Harness: --batch loses its upper bound, the fixture and paged pool are
  sized by it, and the host reference runs decode_fwd_layers once per
  window over absolute block-table page ids.

Separately, migrate the three compile_for_test call sites to the public
compile() API. pypto #2230 removed JITFunction.compile_for_test, which
breaks rope_qkv_regen on a2a3 and decode_layer_a8w8 on the sim sweep on
main as well as here. compile(), not lower(): the removed method ran the
full pipeline and produced artifacts, and these callers depend on that --
rope_qkv_regen's --emit-header re-wraps the generated kernels/aiv file and
the two compile-only sweeps exist to catch codegen regressions, whereas
lower() is codegen-free and artifact-free by contract.

Validated on a2a3: batch 1/8/16 unchanged bit-for-bit against the
previous single-window path, and 17/32/33/48 match the window-wise
reference (logits 100% within 5e-2); batch 32 also passes the full
40-layer stack. Weights are re-read per window, so cost is linear in the
window count: 40 layers measure 33.4 ms at batch 16 and 67.0 ms at 32.

### DIFF
--- a/docs/pypto-coding/cce-extern-kernel.md
+++ b/docs/pypto-coding/cce-extern-kernel.md
@@ -50,13 +50,14 @@ first — several traps below follow directly from it.
 is **not** positional to the Python signature.
 
 For the fused Qwen extern, the declaration starts with the returned attention
-buffer and ends with the scalar:
+buffer and ends with the scalars:
 
 ```python
 def paged_attention_rope_cce(
     out, query, key_cache, value_cache, block_table, workspace, metadata,
     q_proj, k_proj, ..., seq_lens,
     cache_row_offset: pl.Scalar[pl.INDEX],
+    batch_offset: pl.Scalar[pl.INDEX],
 ) -> ...
 ```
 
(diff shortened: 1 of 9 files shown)


lyfne123:
@coderabbitai thanks — dispositions for the three nitpicks, all at `eec7507`.

- **fai_body.hpp, pointer-arithmetic units** — done. The block-table displacement
  is now a named `block_table_byte_offset` alongside the existing
  `cache_byte_offset`, and both sites carry an explicit BYTES/ELEMENTS note:
  `GM_ADDR` displacements are byte counts, while the typed `__gm__ int32_t *`
  arithmetic in the RoPE call advances elements (one entry per row).

- **decode_fwd.py, name-derived window indices** — done. The literal
  `_win[6]/_win[7]/_win[8]` are replaced by a `_slot(name)` helper that derives
  the position from `INPUT_NAMES` (minus one, since `stacked` drops
  `hidden_states`). Verified it resolves to the same 6/7/8 today, so reordering
  `INPUT_NAMES` now moves the windowed tensors with it instead of silently
  windowing the wrong ones.

- **contract.py, cap decode compile dummies at the padding width** — declined for
  now. The current sizing is correct, just larger than strictly needed, and the
  clamp changes the shapes the serving consumer compiles against. I can't
  exercise that compile path here to confirm the dynamic batch axis still binds
  as intended, so I'd rather not take it blind in this PR. Worth doing as its own
  change, validated together with pypto-serving.

Re-validated on a2a3 after the edits: batch 17 and 32 both match the window-wise
reference, logits 100% within 5e-2.

PR #836, opened 2026-07-25
WIP: dsv4-pro Hybrid MXFP8–MXFP4 operator rewrite

yanghaoran29:
## Summary

WIP：DeepSeek V4-Pro 算子按 Hybrid MXFP8–MXFP4 接线（LeftScale `AND2ZZ`、tiled `MX_B_NN`、Acc peel/unroll 等）。  
依赖尚未合入 main 的 pypto MX DSL 与 simpler FP8/FP4 host dtype。

## 依赖版本（本 PR 验证环境）

| 组件 | 版本 / commit | 说明 |
|------|----------------|------|
| **pypto** | `88d2d57e9c84` — [hw-native-sys/pypto#2117](https://github.com/hw-native-sys/pypto/pull/2117) (`issue-1975-mx-dsl-ops`) | MXFP8/MXFP4 tile DSL + Acc prologue dominate + FP4 `reinterpret_view` |
| **simpler**（pypto `runtime` submodule） | `d55ad749f1c6` — [hw-native-sys/simpler#1473](https://github.com/hw-native-sys/simpler/pull/1473) (`add-mx-fp8-fp4-dtypes`) | FP8E4M3FN / E5M2 / E8M0 / FP4E2M1 host dtype |
| **pto-isa** | `83d01313d9bfc247c4b7c8bcf969d1019f0d106f` | 来自 `runtime/pto_isa.pin`（与 simpler 构建一致） |
| **ptoas** | **0.48**（x86_64，`PTOAS_ROOT=~/opt/ptoas-x86_64`） | 本机 `ptoas --version` |
| **本 PR 分支 tip** | `bbb5e1cdd760` — `wip/v4-pro-mxfp8-mxfp4` | |
| 平台 | Ascend A5，device 0；CANN `cann-9.1.0-beta.1` | |

## 上板 golden 状态（A5）

主 sweep：`~/board_logs/sweep_20260726_022600/`（2026-07-26 02:26）。  
其后单独复跑：`qkv_proj_rope`（**RoPE 4-row gather 修复后 decode+prefill 全 PASS**）。

### PASS（整算子 golden 全过）— 7

| 算子 | 备注 |
|------|------|
| `expert_shared` | MXFP8 W8A8，`sh` golden PASS |
| **`qkv_proj_rope`** | decode + prefill：`q`/`kv`/`qr`/`qr_scale` 全 PASS（A5 `tgather` 8 行末行损坏 → RoPE gather 拆成 4 行子块） |
[...]

PR #897, opened 2026-08-05
Fix: align expert_routed golden with kernel int32 matmul order

lwDavid:
## Summary

[#847](https://github.com/hw-native-sys/pypto-lib/pull/847) made `golden_expert_shared` mirror the kernel's int32 accumulate. Its literal sibling in the same file family, `golden_expert_routed`, was left on the old form. This applies the same change.

The kernel does an exact INT8 × INT8 → INT32 accumulate for all three matmuls (`pl.matmul(..., out_dtype=pl.INT32)` + `pl.matmul_acc`) and dequantizes **once, after** the reduction, by per-row input scale × per-channel weight scale. The golden pre-dequantized `w1`/`w3`/`w2` to fp32 and ran fp32 matmuls, accumulating rounded products. Integer accumulation is exact and order-independent, so the int32 form is bit-identical to the kernel's tiled accumulate regardless of K-tiling.

## Two details worth reviewing

**Parenthesisation is load-bearing.** The w2 stage keeps `(h_sd * w_per_row)` grouped, because the kernel forms `row_scale_blk = pl.mul(h_tile_scale_dq, w_col_blk)` as one fp32 product before applying it through `row_expand_mul`. Ungrouping it changes the rounding.

**Host memory.** `dequant_w` materialized all three weights as fp32 tensors held live across the whole expert loop:

```
3 × 48 × 3072 × 7168 × 4 B = 11.81 GiB    (N_LOCAL_EXPERTS=48, MOE_INTER=3072, D=7168)
```

The replacement casts one expert's weights to int32 inside the loop — 84 MiB, freed each iteration. This file has already been the subject of a host-memory fix (#745 on `gen_routed_weight`), so the cost is a known concern here.

## Known limitation, same as #847

The w2 stage still cannot be bit-exact: the kernel computes silu as `recip(add(exp(neg(gate)), 1.0))`, which won't match `F.silu` to the last ULP, and that flips ties in the INT8 requant. Only w1/w3 become exact.

## Scope

Golden only. No kernel change, no device-side change, no API change.

Note for @yanghaoran29: draft #836 rewrites this golden wholesale for MXFP8 and does not include the int32 alignment — worth picking up there.

### DIFF
--- a/models/deepseek_v4_pro/expert_routed.py
+++ b/models/deepseek_v4_pro/expert_routed.py
@@ -276,16 +276,24 @@ def golden_expert_routed(tensors):
     import torch
     import torch.nn.functional as F
 
-    def dequant_w(w_i8, w_scale):
-        return w_i8.to(torch.float32) * w_scale.unsqueeze(-1)
-
+    # Mirror the kernel's numerics exactly, as golden_expert_shared already does.
+    # Every matmul in the kernel is an exact INT8 x INT8 -> INT32 accumulate,
+    # dequantized AFTER the reduction by the per-row input scale x per-channel
+    # weight scale. Integer accumulation is exact and order-independent, so an
+    # int32 matmul here is bit-identical to the kernel's tiled accumulate
+    # regardless of K-tiling. A fp32 matmul of pre-dequantized operands (the
+    # previous form) instead accumulates fp32 rounding error that flips bf16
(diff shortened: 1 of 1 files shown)


lwDavid (commented):
Reviewed. The golden computation is aligned with device datatype. The quant method should be changed after more datatypes are supported.

Issue #136, opened 2026-04-21
[Feature] [Tracking] PagedAttention support for Qwen3 inference (multi-request / serving)


lyfne123:
### Summary

Add **PagedAttention** support to the Qwen3 inference path, enabling multi-request concurrent decoding, long-context efficiency, and integration with serving frameworks (vLLM / SGLang / MindIE). This is the **Phase 2** follow-up to the Qwen3-14B end-to-end bring-up tracked in #135.

Detailed design captured in `examples/models/qwen3/14b/qwen3_14b_e2e_guide.md` §6 (Phase E).

### Motivation / Use Case

The current single-layer kernel (`examples/models/qwen3/qwen3_32b_decode.py`) and the Qwen3-14B end-to-end work in #135 both use a **contiguous KV cache** layout `[B × H_kv × MAX_SEQ, D]`. This is fine for single-request demos but **cannot** support:

| Scenario | Contiguous KV | Paged KV |
|---|---|---|
| Single-request demo | ✅ | ✅ |
| Multi-request concurrency (varying seq_len) | ❌ wastes memory by `MAX_SEQ` per request | ✅ block-level allocation |
| Long context (>32K) with high concurrency | ❌ each request eats GBs | ✅ on-demand blocks |
| Prefix caching / shared system prompts | ❌ | ✅ |
| Beam search / parallel sampling | ❌ | ✅ (block refcounting) |
| Integration with vLLM / SGLang / MindIE | ❌ interface mismatch | ✅ standard contract |

PagedAttention is the **de-facto industry standard** for LLM serving and is required for any production-style deployment.

#### Kernel-side changes

Decode / prefill attention kernels accept new inputs:

```python
@pl.function(...)
def qwen3_decode_paged(
    self,
    hidden_states: pl.Tensor[[batch, hidden], pl.BF16],
    # ... weights, RMSNorm γ, Q/K norm, RoPE tables ...
    seq_lens:     pl.Tensor[[batch], pl.INT32],
    block_table:  pl.Tensor[[batch, max_blocks_per_seq], pl.INT32],   # logical → physical block
    slot_mapping: pl.Tensor[[batch], pl.INT32],                       # write slot for new token
    k_cache: pl.Tensor[[num_blocks, block_size, num_kv_heads, head_dim], pl.BF16],
    v_cache: pl.Tensor[[num_blocks, block_size, num_kv_heads, head_dim], pl.BF16],
[...]

PR #902, opened 2026-08-05
Chore: CI probe for a serving-qwen baseline on unmodified main

lyfne123:
DO NOT MERGE — throwaway diagnostic, will be closed once the job reports.

This branch is `main` plus one comment line under `models/qwen3_14b/`, which
exists only so `detect-changes` schedules the `serving-qwen` job.

`serving-qwen` has not executed the model in recent history. Every non-skipped
run either hit the runner's `simpler_init` 507018 or, after #886 flattened
`models/`, the pypto-serving path lookup. pypto-serving #136 fixed that lookup,
and the first run to get past it — on #892 — failed with all-zero output tokens
from `test_qwen3_output_matches_expected_tokens`. There is no green run to
attribute that to.

Its `serving-qwen` result is that missing baseline: if this fails the same way,
the breakage predates #892; if it passes, #892 is implicated. The other jobs on
this branch are expected to fail on `compile_for_test` (removed by pypto #2230,
still unfixed on main) and are not what this probe is for.

### DIFF
--- a/models/qwen3_14b/decode_fwd.py
+++ b/models/qwen3_14b/decode_fwd.py
@@ -7,6 +7,7 @@
 # See LICENSE in the root of the software repository for the full text of the License.
 # -----------------------------------------------------------------------------------------------------------
 # ci: no-sim    # CI marker: external CCEC attention is A2/A3 onboard only
+# CI PROBE -- comment-only touch so detect-changes runs serving-qwen. Do not merge.
 """Qwen3-14B decode with FP32 inter-layer carry and direct CANN attention.
 
 The projection, QK-norm, RoPE, output projection, MLP, and dependency topology


lyfne123:
Probe done — result recorded.

This branch is `main` plus one comment, and its `serving-qwen` reproduced #892's failure byte for byte:

    expected token_ids: [12095, 13, 3555, 374, 279, 6722, 315, 279]
    assert [0, 0, 0, 0, 0, 0, ...] == ...

So the all-zero decode output predates #892 and is not caused by it. The qwen serving path has been producing zeros on main while `serving-qwen` was masked by earlier failures (the runner's 507018, then the post-#886 pypto-serving path lookup that #136 fixed).

Note the green `a2a3`/`sim` here is not evidence about `compile_for_test`: the model sweep only runs CHANGED files, and this branch changed only decode_fwd.py, so rope_qkv_regen.py and decode_layer_a8w8.py never ran. pypto main still has no `compile_for_test`.

Closing; not for merge.

Issue #905, opened 2026-08-06
[Feature] DeepSeek-V4-Flash serving directory targeting the vllm-ascend A3 single-node DSpark point: DP4 TP4 EP16 + CP, B=64 S=8, 1M context, W8A8


zhangqi-chen:
### Summary

Bring `models/deepseek_v4_flash_dspark/` to the **vllm-ascend A3 single-node DSpark
deployment point**: one Atlas 800 A3 node (16 die) running Prefill and Decode
together at **DP4 × TP4 ⇒ EP16 with DSA CP**, `B = 64`, `S = 8`, 1M context, W8A8.

**References** — everything below is derived against these three:

| | |
|---|---|
| Deployment doc | <https://docs.vllm.ai/projects/ascend/en/latest/tutorials/models/DeepSeek-V4-Flash.html> — §5.1 tab *"A3 series DSpark"*, §9.1 (zh-cn: <https://docs.vllm.ai/projects/ascend/zh-cn/latest/tutorials/models/DeepSeek-V4-Flash.html>) |
| Code | <https://github.com/vllm-project/vllm-ascend> at **`ac19e1e`** — every `vllm_ascend/…:line` citation is read there |
| Weights | `DeepSeek-V4-Flash-0731-w8a8` (`dspark_block_size: 5`, `dspark_target_layer_ids: [40,41,42]`, `dspark_markov_rank: 256`, `dspark_noise_token_id: 128799`) |

    vllm serve /path/to/DeepSeek-V4-Flash-0731-w8a8 \
      --max-model-len 1048576  --max-num-batched-tokens 10240  --max-num-seqs 64 \
      --data-parallel-size 4   --tensor-parallel-size 4        --enable-expert-parallel \
      --block-size 32          --quantization ascend \
      --speculative-config '{"method":"dspark","num_speculative_tokens":7,"enforce_eager":true}' \
      --compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}' \
      --additional-config '{"enable_dsa_cp": true, "enable_flashcomm1": true,
                            "multistream_overlap_shared_expert": true, "enable_cpu_binding": true}'

Seven adaptation items:

| | Item | State |
|---|---|---|
| P1 | S = 8, batch 64, and the two per-die widths | partly done — #907 |
| P2 | Decode context to 1M, block size 32 | open |
| P3 | **Decode parallelism and communication** | open |
| P4 | Maximum prefill input | open |
[...]

PR #913, opened 2026-08-07
Add: experimental shared-expert weight TP on fixed SP

wangqin1723-max:
Part of #905.

- Keep shared-expert decode on fixed SP4-local token shards, with replicated weights and no shared-expert TP collective by default.
- At the merged S=8 configuration, each SP rank owns 128 of the global 512 decode rows.
- Add an explicit experimental `--tp-shared-expert {1,2,4}` degree instead of inheriting the model-wide TP setting.
- Shard only the shared intermediate weights for experimental TP2/TP4 groups while leaving routed-MoE token shapes and EP dispatch/combine unchanged.
- Leave prefill behavior unchanged.

Validation after rebasing onto main at `d1cf017` (S=8, block size 32):

- `pre-commit run --all-files`
- Standalone shared expert: full A2/A3sim runtime + golden for TP1, TP2, and TP4
- Integrated MoE: compile for EP2/TP1, EP2/TP2, and EP4/TP4
- Pin-aligned pre-rebase CI A2/A3sim: standalone shared expert and integrated EP2/TP1 both PASS
- Physical A3 before and after the latest-main rebase: standalone shared expert and integrated EP2/TP1 both PASS; `x_next` shape `[2, 128, 4, 4096]`
- Integrated EP2/TP1 with level-4 L2 instrumentation: physical A3 runtime + golden PASS

Known simulator limitations:

- The post-rebase A2/A3sim job passed standalone `expert_shared.py`, then the integrated `moe.py` scheduler stalled at task `4294967315`; the workflow canceled it at 30 minutes. A different local simulator stack can instead stall at the existing early-dispatched `combine_wait`, while level-4 instrumentation changes timing and passes. This is scheduler timing sensitivity, not a numerical mismatch.
- The standalone shared expert also passes A5sim. Fresh integrated `moe.py` runs produce roughly 2.1 million NaNs in `x_next` (2,162,688 before the rebase and 2,129,920 after it).
- A pin-identical replay of upstream main `2fdc9fc` with the same saved inputs returns exactly the same 2,162,688 NaNs at the PR head.
- The first isolated invalid boundary is unchanged `hc_pre.py`: `x_mixed` and `post` pass, while `comb` contains 1,024 NaNs out of 2,048 values on A5sim. Enabling dependency capture does not change the result, and the generated dependency graph already contains the producer-to-`comb_sinkhorn` edges.
- Neither simulator result is a shared-expert TP regression. The PR intentionally does not widen numerical tolerances or modify unrelated HC/MoE scheduling to hide them.


wangqin1723-max:
CI triage for the remaining simulator results:

- Post-rebase A2/A3sim passed standalone `expert_shared.py`, then integrated `moe.py` hit scheduler stall task `4294967315`; the workflow canceled it at 30 minutes. The same patch had passed the prior pin-aligned A2/A3sim CI and still passes real A3, so this is timing-sensitive simulator scheduling.

- `expert_shared.py` compiles, runs, and passes its BF16 golden check on A5sim.
- Fresh integrated `moe.py` runs produce roughly 2.1 million NaNs in `x_next` (2,162,688 before the rebase and 2,129,920 in the post-rebase CI run).
- With PyPTO `d315f62`, simpler `3165cc8`, PTOAS v0.54, and PTO ISA `83d0131`, upstream main `2fdc9fc` returns exactly the same 2,162,688 NaNs. Replaying the same saved inputs at PR head `6b02a4c` gives the identical result.
- The first isolated invalid boundary is unchanged `hc_pre.py`: `x_mixed` and `post` pass, while `comb` contains 1,024/2,048 NaNs on A5sim. Every 8-row tile has NaNs in rows 4–7 across all 16 columns.
- `enable_dep_gen=True` does not fix the result, and the captured graph already has the required producer edges into `comb_sinkhorn`.
- The current PR passes A2/A3sim and physical A3, including integrated `moe.py` output `[2, 128, 4, 4096]`.

This is a pre-existing A5-specific `hc_pre/comb_sinkhorn` kernel or lowering problem, not a shared-expert TP regression. I am not widening tolerances or changing unrelated scheduler/kernel code to mask it.

models/deepseek_v4_flash_dspark/expert_shared.py
```diff
@@ -6,81 +6,269 @@
 # INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
 # See LICENSE in the root of the software repository for the full text of the License.
 # -----------------------------------------------------------------------------------------------------------
-"""DeepSeek-V4 MoE shared expert compute (decode, EP single-card).
+# ci: devices=4
```
zhangqi-chen, line 9: default ci =1

PR #925, opened 2026-08-10
Add: support decode attention TP and SP boundaries

wangqin1723-max:
- Gather contiguous SP token shards before decode attention and
  reduce-scatter FP32 O-B partials back to their owning ranks.
- Shard Q-B heads, attention sinks, O-A groups, and O-B inputs for
  SWA, HCA, and CSA decode at selectable TP1/TP2/TP4/TP8 degrees.
- Select the single-device L2 JIT entry for TP1 and the distributed L3
  host wrapper for TP2/TP4/TP8. TP1 creates no DistributedConfig and
  runs no SP AllGather or ReduceScatter.
- Size decode metadata, compressors, and indexer paths for the native
  64-request attention batch instead of the pre-TP batch stand-in.
- Preserve the full-head QKV entry used by unchanged prefill callers
  while decode TP uses a separate rank-local QKV entry.
- Order collective completion and signal reset before downstream
  consumers so communication windows are safe to reuse across layers.

### DIFF
--- a/models/deepseek_v4_flash_dspark/attention_tp.py
+++ b/models/deepseek_v4_flash_dspark/attention_tp.py
@@ -1,0 +1,246 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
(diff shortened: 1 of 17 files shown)


wangqin1723-max:
Issue #905 now targets the vllm-ascend DSA-CP-on layout. This PR models the previous CP-off direction by sharding Q-B/attention heads and gathering tokens before attention, which conflicts with the updated contract: Q-B and attention sink must be replicated, each rank keeps its local 128 queries over all 64 heads, and attention exits through a group-major all-to-all before sharded O-A/O-B plus reduce-scatter.

The replacement is split into reviewable decode-only slices. The first draft is #931, which defines the three communication seams and exposes the SWA packed-head/task dependency boundary. HCA/CSA and the sharded receive-side O projection will follow. Prefill remains unchanged. Closing this PR as superseded; its history stays available for reference.

PR #930, opened 2026-08-10
Perf: shard rms_norm over D and halve the proj_b D-chunk

lwDavid:
- rms_norm: split the decode grid over D. A decode step has
  t_dim == T_TILE, so the SPMD was ONE block and the whole row was
  normalized on a single core while it sat on the critical path. Each
  block now recomputes the same full-row sum of squares and applies only
  D // RMS_D_SPLIT columns, which keeps the task one dispatch with no
  cross-block reduction. The token axis cannot go finer: T_TILE < 8 puts
  the [1, T_TILE] FP32 row accumulator below the 32-byte alloc-tile row
  floor and codegen rejects it.
- decode_sparse_attn: halve PROJ_B_D_CHUNK to 1792. The 3584 chunk was
  sized so its 32 cube blocks form one wave over the 28 cube cores, but
  proj_b bundle g is gated by its own proj_a -> quant chain, so the eight
  bundles enter the array staggered and the array drains at the end of
  the tail. Half-width chunks halve the quantum a late bundle has to fit
  into that drain. 1024 was swept too and gave the win back.

Both are bit-identical: every rms_norm block derives x_inv_rms from the
same inputs in the same chunk order and writes disjoint columns, and
proj_b's K-reduction per output column is unchanged.

CSA decode latency 716.9 -> 704.7 us (a5 decode, PYPTO_BENCH 100 rounds,
op fusion disabled). Measured separately the two are worth -5.9 us and
-6.1 us.

### DIFF
--- a/models/deepseek_v4_pro/decode_sparse_attn.py
+++ b/models/deepseek_v4_pro/decode_sparse_attn.py
@@ -141,10 +141,19 @@ PA_BLOCKS = PA_NFRAGS // PA_NF_PER_BLOCK
 assert PA_NFRAGS % PA_NF_PER_BLOCK == 0, "proj_a N-frag loop must cover PA_NFRAGS"
 # proj_b is one task per (D-chunk, group): the D-chunk's N-frags loop INSIDE the task,
 # so the per-group split does not multiply the task count by N-frags. The block count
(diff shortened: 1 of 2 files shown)


lwDavid (commented):
Reviewed. Shard rms_norm into 4 blocks along D and halve the CSA output projection's PROJ_B_D_CHUNK to 1792.

PR #931, opened 2026-08-10
Add: define decode DSA-CP communication and output seams

wangqin1723-max:
Part of #905. This is the first replacement slice for the old CP-off attention direction in #925.

## What

- Define the TP4/SP4 decode component layout: Q-B, attention sink, and shared expert replicated; O-A, O-B, and vocabulary sharded four ways.
- Add subgroup-local hidden-state all-gather, group-major attention-output all-to-all, and FP32 O-B reduce-scatter seams.
- Make all three collectives capacity-static but runtime-row aware: valid rank slabs are packed at `rank * local_t`, and invalid tails remain untouched.
- Add the receive-side sharded O projection: two local O-A groups, a 2048-wide O-B K shard, per-group A8 quantization, and an FP32 partial for reduce-scatter.
- Split the SWA, HCA, and CSA decode kernels at the same group-major packed-head/task-ID boundary while preserving each legacy public wrapper and local output projection.
- Use simulator-compatible single-buffer TPUT transfers with an 8-row communication tile.

## Layout contract

Each attention core publishes `[group, T_PAD, head-in-group, dim]`; only the first runtime token rows in each group are valid. The all-to-all writes each source rank directly into `[local_group, source_rank * local_t]`, producing two local O-A groups over the full TP-group token stream.

The receive-side projection consumes `[LOCAL_O_GROUPS, GROUP_T_PAD, O_GROUP_IN]`, computes sharded O-A and O-B over only `group_t = SP_SIZE * local_t`, and returns a rebound `[GROUP_T_PAD, hidden]` FP32 partial plus its completion task ID. Reduce-scatter then sends each token owner compact `local_t` rows and casts only the reduced local result to BF16.

## Validation\n\n- Rebased without patch changes onto main at `d1cf017` (#932)\n- `pre-commit run --all-files`
- Four-rank A3 simulator + exact golden for AG/A2A/RS at current S=2:
  - full capacity `local_t=32`
  - poisoned subcapacity `local_t=31`, including preserved output tails
- Merged #928 S=8 base: four-rank A3 simulator + exact golden at full capacity `local_t=128` and dynamic subcapacity `local_t=127`
- Receive-side O projection A3 simulator + golden:
  - current S=2 at 128/124 group rows
  - merged #928 S=8 base at 512/508 group rows
  - strict valid-prefix check with NaN-poisoned receive padding and zero output tails
  - group-distinct O-A weights spanning all 4096 K columns and non-unit O-B scales
- SWA seam: A3 simulator + golden at current local T=32 and temporary S=8 local T=128
- HCA seam: A3 simulator + golden at T=32, subcapacity T=8, and temporary S=8 T=128
- CSA seam: T=32, subcapacity T=8, and temporary S=8 T=128 compile and complete A3 simulator runtime. The pinned local stack has a pre-existing broad CSA golden mismatch reproduced on the untouched base; after normalizing renamed symbols, all nine generated PTO kernels are byte-identical to the base.

Physical A3 CI passed the four-rank communication fixture, the strict 512/508-row receive-side O projection cases, all three staged sparse-attention programs, and the unchanged LM-head, including a fresh pass after rebasing onto `d1cf017`. The first A3 job had a transient unchanged LM-head S1 scheduler stall; two no-code-change A3 runs passed. Because `config.py` changes, simulator CI swept all 32 runnable files in the directory: both platforms passed the new communication and O-projection fixtures, then reproduced existing directory-wide mismatches (including the CSA mismatch reproduced on the untouched base with byte-identical generated kernels) and eventually stalled in unchanged `lm_head.py` until the 30-minute cancellation. The red simulator statuses are therefore not feature-local regressions. No physical A5 run was available. Prefill is unchanged by this PR. SWA, HCA, and CSA end-to-end output-path composition follow as decode-only slices; #913 and #923 cover the shared-expert and vocabulary boundaries.\n

### DIFF
--- a/models/deepseek_v4_flash_dspark/config.py
+++ b/models/deepseek_v4_flash_dspark/config.py
@@ -269,16 +269,17 @@ FP32_NEG_INF = -3.4028234663852886e38     # most-negative finite fp32 (softmax m
 
 # Parallelism constants
 TP = 4         # tensor-parallel ranks per DP group
+SP = TP        # sequence-parallel token owners in the TP group
 DP = 4         # DP groups per node
 EP = 16        # expert-parallel world size (moe overrides it from --ep)
 
 # Per-component TP degree, over the components that shard.
-TP_Q_B = TP            # wq_b: ColumnParallel over heads
-TP_O_A = TP            # wo_a: ColumnParallel over o_groups
-TP_O_B = TP            # wo_b: RowParallel, then allreduce
-TP_ATTN_SINK = TP      # attn_sink: one entry per local head
-TP_SHARED_EXPERT = TP  # shared expert: gate_up ColumnParallel, down RowParallel then allreduce
-TP_VOCAB = TP          # embed_tokens / lm_head: vocab-parallel
+TP_Q_B = 1            # wq_b: replicated across the DSA-CP group
+TP_O_A = TP           # wo_a: ColumnParallel over o_groups
+TP_O_B = TP           # wo_b: RowParallel, then reduce-scatter
+TP_ATTN_SINK = 1      # attn_sink: all attention heads on every DSA-CP rank
+TP_SHARED_EXPERT = 1  # shared expert: sequence-parallel with replicated weights
+TP_VOCAB = TP         # embed_tokens / lm_head: vocab-parallel
 
 # MoE constants
 MOE_TOKENS = DECODE_TOKENS * DP // EP
--- a/models/deepseek_v4_flash_dspark/decode_o_projection_cp.py
+++ b/models/deepseek_v4_flash_dspark/decode_o_projection_cp.py
@@ -183,6 +183,8 @@ def build_tensor_specs(local_t):
 
     from golden import ScalarSpec, TensorSpec
 
+    if local_t < 1 or local_t > LOCAL_T:
+        raise ValueError(f"local_t must be in [1, {LOCAL_T}], got {local_t}")
     group_t = SP_SIZE * local_t
 
     def init_attention_local_groups():
--- a/models/deepseek_v4_flash_dspark/decode_o_projection_cp.py
+++ b/models/deepseek_v4_flash_dspark/decode_o_projection_cp.py
@@ -1,0 +1,285 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


models/deepseek_v4_flash_dspark/decode_o_projection_cp.py
```diff
@@ -0,0 +1,285 @@
...
+    return o_partial
+
+
+def build_tensor_specs(local_t):
+    """Build deterministic capacity-static inputs with a poisoned receive tail."""
+    import torch
+
+    from golden import ScalarSpec, TensorSpec
+
+    group_t = SP_SIZE * local_t
```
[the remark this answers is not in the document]
wangqin1723-max: Fixed in 2f1601f: the output-projection fixture now rejects local_t outside [1, LOCAL_T]. I also made the poisoned attention tail return explicit and added fixed BF16 layout annotations at the CSA/HCA/SWA packed-head seams. Targeted compile-only validation and pre-commit pass.

wangqin1723-max:
Latest-main CI disposition for head `d5f894d` on `main@d1cf017`:

- Pre-commit, unit tests, build, and physical A3 all pass.
- The A3 job passed the new four-rank communication fixture, strict 512/508-row O projection, and all CSA/HCA/SWA staged programs.
- Updating `config.py` makes simulator CI sweep all 32 runnable dspark files. Both simulators passed `decode_attention_cp.py` and `decode_o_projection_cp.py`, then reproduced existing directory-wide mismatches. The changed CSA wrapper has the same broad mismatch reproduced on untouched main, with all nine generated PTO kernels byte-identical after symbol normalization.
- Both simulator jobs eventually stalled in unchanged `lm_head.py` and were canceled at the 30-minute workflow limit.

I am treating the simulator statuses as baseline/runtime debt, not changing DSA-CP kernels or tolerances to mask them.

models/deepseek_v4_flash_dspark/config.py
```diff
@@ -269,16 +269,17 @@ def mix_hc(self) -> int:
 
 # Parallelism constants
 TP = 4         # tensor-parallel ranks per DP group
+SP = TP        # sequence-parallel token owners in the TP group
```
zhangqi-chen, line 272: delete sp

models/deepseek_v4_flash_dspark/decode_o_projection_cp.py
```diff
@@ -0,0 +1,287 @@
+# Copyright (c) PyPTO Contributors.
```
zhangqi-chen, line 1: in next pr, rename this to decode_o_proj.py，put all o proj related functions within it, including communication, tp=1 functions.

PR #936, opened 2026-08-11
Fix: stabilize DeepSeek V4 Pro prefill CSA on A5

lwDavid:
- Keep prefill RoPE gather indices row-sized in QKV projection and both
  ratio-4 compressors to avoid wide A5 TCI tiles overrunning UB
- Defer Q dequantization until Q projection finishes instead of reserving
  AIV workers early


lwDavid (commented):
Reviewed. Make RoPE gathers row-wise and adjusting Q-projection scheduling.

PR #942, opened 2026-08-11
Fix: build qwen3 paged-attention CCE against both runtime ABIs

Little-oil:
- Add runtime_tensor_compat.hpp, which probes for the runtime-only
  task_interface/buffer.h header and aliases PyPTORuntimeTensor to
  ChipTensor when that header exists and to Tensor otherwise.
- Cast through PyPTORuntimeTensor at the two runtime-descriptor access
  points, in the FAI kernel body and in the paged-attention tiling
  entry, replacing their direct "tensor.h" includes.
- Keep reading tensor data as buffer.addr + start_offset, so argument
  layout, offsets, and device-address calculation are unchanged on the
  AIC, AIV, RoPE, and non-RoPE paths.

Simpler renamed the device-side Tensor descriptor to ChipTensor in its
address-free Buffer ABI. The alias lets pypto-lib compile against both
that runtime and the currently pinned one, so this change can land
before hw-native-sys/pypto#2273 moves the runtime pin; it can be
dropped once the legacy runtime is no longer supported.

### DIFF
--- a/models/qwen3_14b/kernels/paged_attention_cce/kernel/fai_body.hpp
+++ b/models/qwen3_14b/kernels/paged_attention_cce/kernel/fai_body.hpp
@@ -28,7 +28,7 @@
 #endif
 
 #include "intrinsic.h"
-#include "tensor.h"
+#include "runtime_tensor_compat.hpp"
 
 #include "../generated/kernel_tiling/kernel_tiling.h"
 #include "metadata_layout.h"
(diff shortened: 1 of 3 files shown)


Little-oil:
@zhangqi-chen 麻烦帮忙 review/merge。该兼容修复是 PyPTO #2273 升级到 Simpler ChipTensor ABI 后 pypto-lib-model CI 的前置依赖；本 PR 全部 CI（含 serving-qwen 上板）已通过。

PR #949, opened 2026-08-12
Fix: make DeepSeek V4 CSA regression deterministic

wangqin1723-max:
- Seed the standalone CSA fixture before constructing its inputs and
  torch golden so the known TPipe-sensitive case is reproducible across
  local and CI runs.

Depends on hw-native-sys/pto-isa#240.

### DIFF
--- a/models/deepseek_v4_flash_mtp/decode_csa.py
+++ b/models/deepseek_v4_flash_mtp/decode_csa.py
@@ -900,6 +900,8 @@ if __name__ == "__main__":
     parser.add_argument("--start-pos", type=int, default=None,
                         help="Uniform fixture-only start_pos override for all batches; "
                              "default (unset) uses the canonical per-batch CSA set that includes the 8k point.")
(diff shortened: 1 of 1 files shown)


wangqin1723-max:
Device validation for the fixed default seed (`1807`) on A2/A3 device 4:

- PTO-ISA `0cefc9a5`: `x_out` fails with about 46% mismatched points.
- PTO-ISA `d2b30a84`: scheduler stalls before validation.
- PTO-ISA #240 head `cae2aac1`: both `kv_cache` and `x_out` pass in 3/3 independent runs.
- Restored control `83d01313`: both outputs pass.

The tested PyPTO revision `36e53454` contains the double-ring workspace support from pypto#2271; PTOAS was fixed at 0.57 and Simpler at `50c06606`.

The downstream result has also been posted on hw-native-sys/pto-isa#240 to request its GitCode/merge integration and a subsequent Simpler pin update.

PR #972, opened 2026-08-17
Add: dspark prefill operators take a runtime token count to 8192

MaxwellF1:
- Take the prefill token count as a runtime tensor extent instead of
  the compile-time PREFILL_SEQ constant. Every prefill operator now
  accepts a token count in [1, PREFILL_MAX_TOKENS] and rejects anything
  outside it, so one compiled program serves every request length
  instead of one variant per length.
- Replace the static [T, ...] signature dimensions with a per-operator
  pl.dynamic token axis, size the intermediates from the tensor extent
  itself, and drop the num_tokens scalar that marked how many rows of a
  fixed 128-row buffer were live.
- Round the token block counts up rather than down and guard the
  trailing block with a row predicate or a valid_shape on the load, so
  a token count that is not a multiple of the block size keeps its
  last rows.
- Split the compressor, indexer, qkv projection, sparse attention and
  rmsnorm bodies into an aligned inner tile plus an outer loop over the
  runtime extent, leaving the aligned path identical to the previous
  fixed-shape dataflow.
- Carry the ratio-4 and ratio-128 compressor pooling state across those
  inner tiles. The state ring holds 520 rows, so the tile is 512 rows,
  one tile plus the eight-row pooling window, and an ordering token
  keeps the next tile from overwriting a row the current one still
  reads.
- Derive the indexer candidate bound per query wave from that wave's
  own position instead of from the request's final position.
- Take the dense projection tile to 512 rows and share one set of rope
  tables across those tiles, so an 8192-token layer issues 16
  projection tiles where a 128-row operator needed 64.
- Add PREFILL_MAX_TOKENS to config; the dense, state and query tile
  sizes stay in the operators that use them. Parameterise each
  operator's tensor specs and CLI by token count.


MaxwellF1:
### Reading the diff

The change is one thing applied to every prefill operator: the token count moves
from a compile-time constant to a runtime extent. Five mechanisms cover the
whole diff.

1. **Dynamic axis** — `B / S / T = 128` gives way to `T_DYN = pl.dynamic(...)`.
   Signature dimensions become `T_DYN` and the intermediates are sized from the
   incoming extent rather than from `PREFILL_SEQ`.
2. **Round up, then guard the tail** — `t_dim // TILE` becomes
   `(t_dim + TILE - 1) // TILE`, with either a row predicate or a `valid_shape`
   on the trailing block. Under the old fixed 128 this case could not arise.
3. **Aligned inner tile** — the previous dataflow moves unchanged into a
   `_*_tile(...)` helper driven by an outer loop over the runtime extent, so the
   aligned path is the same code it was before. That is why the existing goldens
   carry over.
4. **`num_tokens` dropped** — length used to be a scalar carried beside a fixed
   128-row buffer; it is now the tensor extent itself.
5. **Specs parameterised by token count** — `build_tensor_specs(token_count=...)`
   plus a `--token-count` flag on every operator.

`hc_post.py` is the smallest complete example of mechanism 2. `prefill_swa.py`
is the smallest of 1 and 4, and is a net line reduction.

### Where the 512-row compressor tile comes from

`CSA_STATE_PHYSICAL_BLOCKS = 260` at `C4A_COMPRESSOR_BLOCK_SIZE = 2` gives a
520-row state ring, and ratio-4 pooling reads back `STATE_LEN = 8` rows.
`520 - 8 = 512` is therefore the largest tile that cannot overwrite a row it
still has to read. Both constants are unchanged from `main`; a larger tile would
mean raising the physical block count first.

### Validation

Every prefill operator passes its own golden on a single card:

| | |
|---|---|
| `prefill_swa` | PASS |
| `prefill_csa` | PASS |
| `prefill_hca` | PASS |
| `prefill_indexer` | PASS |
| `prefill_indexer_compressor` | PASS |
| `prefill_sparse_attn` | PASS |
| `prefill_compressor_ratio4` | PASS |
| `prefill_compressor_ratio128` | PASS |
| `qkv_proj_rope` | PASS |
| `hc_pre` / `hc_post` / `hc_head` | PASS |
| `rmsnorm` | PASS |

Length sweep, each point running that operator's own golden:

| operator | T=1 | T=129 | T=257 | T=512 | T=8192 |
|---|---|---|---|---|---|
| `prefill_indexer` | PASS | PASS | PASS | PASS | PASS |
| `prefill_csa` | PASS | PASS | PASS | PASS | PASS |
| `prefill_compressor_ratio4` | PASS | PASS | PASS | PASS | PASS |

`T=129` and `T=257` cross a block boundary at a non-multiple, exercising
mechanism 2. `T=512` sits exactly on the compressor state tile, the boundary the
520-row ring constrains.

### Scope

Operators only, matching what this directory already contains — there is no full
prefill forward here. Assembling these into one is separate work and is not part
of this PR, and no performance claim is made.

models/deepseek_v4_flash_dspark/config.py
```diff
@@ -247,6 +247,8 @@ def mix_hc(self) -> int:
 PREFILL_BATCH = 1                 # B: prefill batch for the current kernel programs
 PREFILL_SEQ = 512                 # S: prefill sequence for the current kernel programs
 PREFILL_TOKENS = PREFILL_BATCH * PREFILL_SEQ
+# Public physical-dynamic prefill limit for Issue #905 P4.
+PREFILL_MAX_TOKENS = 8192
```
zhangqi-chen: delete this, but make sure 8k is ok

models/deepseek_v4_flash_dspark/rmsnorm.py
```diff
@@ -39,26 +66,47 @@ def rms_norm(
...
-            norm_w_chunk = pl.cast(pl.reshape(norm_w[apply_d0 : apply_d0 + D_TILE], [1, D_TILE]), pl.FP32)
-            x_normed_chunk = pl.col_expand_mul(pl.row_expand_mul(apply_x_chunk, x_inv_rms_t), norm_w_chunk)
-            x_normed[tg : tg + T_TILE, apply_d0 : apply_d0 + D_TILE] = pl.cast(
-                x_normed_chunk,
-                target_type=pl.BF16,
-                mode="rint",
-            )
+        valid_rows = pl.min(T_TILE, t_dim - tg)
+        if valid_rows == T_TILE:
+            _rms_norm_full_tile(x, norm_w, x_normed)
```
zhangqi-chen, line 133: Replace this with full code, or extract the else branch to another inline function. check other files with same pattern.

zhangqi-chen (changes requested):

zhangqi-chen (approved):

PR #976, opened 2026-08-18
Add: enable elastic 1M decode attention

### LINKED ISSUE #962: [Feature][P2] Enable elastic 1M-context decode attention with block size 32
## Parent tracking issue

- Parent: #905
- Parent work item: **P2 — Decode context to 1M, block size 32**

## Summary

Enable elastic 1M-context decode for `models/deepseek_v4_flash_dspark/` while preserving the existing DeepSeek-V4-Flash attention math and the deployment point defined by #905 (`B=64`, `S=8`, block size 32).

Here, **1M is a capacity ceiling, not a fixed execution shape**. Runtime work must follow the actual visible context of each request. Short requests must not allocate or submit 1M-sized attention/indexer work, and 128/16K/1M must not become separate kernel profiles.

## Goals

1. Make block size 32 the canonical decode cache ABI.
2. Support per-request logical positions and cache capacity up to 1,048,576 tokens.
3. Preserve heterogeneous and inactive requests through ragged runtime metadata.
4. Enable all three decode attention modes:
   - SWA / ratio 0
   - HCA / ratio 128
   - CSA / ratio 4 with exact Top-512
5. Keep production attention structure as close to the existing baseline as possible; long-context support should primarily change metadata, persistent-cache addressing, and the amount of submitted work.

### Common geometry and metadata

- Canonical capacity: `MAX_CONTEXT_TOKENS = 1_048_576`.
- Cache storage block size: 32 rows.
- Use allocator-owned global physical pools; do not pre-partition cache capacity by a fixed batch.
- Device entries consume ragged descriptors containing physical page IDs, valid ranges, epochs, write slots, and exact packed work.
- Page IDs and write slots are local to one layer's pool view and must fail closed when invalid, stale, inactive, or out of range.
- RoPE inputs are token/event-local. Do not pass a `[1M, ...]` RoPE table.
- Runtime length changes task counts and tail validity only; it must not change kernel semantics or select a length-specific profile.

### SWA (ratio 0)

- Semantic window remains 128 rows, stored as four 32-row pages per admitted request.
- `swa_sources[Q, 128]` represents persistent rows, invalid rows, and current-step overlay rows.
- Current-step KV participates causally through the overlay and is committed to the persistent ring only after attention has finished reading the old rows.
- Contexts from 128 through 1M use the same 128-row attention tile and persistent capacity.

### HCA (ratio 128)

- A 1M request produces at most 8,192 compressed rows.
- Attention uses fixed 128-row shards; the runtime submits 0–64 shards per query from exact packed work.
- The ratio-128 state ring retains 128 semantic rows using sixteen 8-row state pages.
- Compressed KV uses ragged 32-row pages and explicit page epochs.
- Boundary compression events use event-local RoPE and explicit main/state write dependencies.
- Raw SWA history plus compressed HCA history are merged with online softmax; no dense history index or `topk_all` tensor is allowed.

### CSA (ratio 4)

- A 1M request produces at most 262,144 candidates.
- Fixed selector leaf: 2,048 candidates -> sorted Top-512.
- Runtime submits only `ceil(valid_candidates / 2048)` active leaves (0–128 per query), followed by a bounded exact merge forest.
- Main KV, index cache, main state, and inner state use independent allocator-owned pools.
- Index cache remains a flat-row ABI; candidate/page work is ragged and exact.
- Preserve the baseline HC/QKV/compressor/sparse-value structure. If micro-chunking is required for the selector forest, confine it to the indexer work instead of repeating the full attention pipeline.

## Production code scope

Primary files:

- `models/deepseek_v4_flash_dspark/config.py`
- `models/deepseek_v4_flash_dspark/context_geometry.py`
- `models/deepseek_v4_flash_dspark/decode_metadata.py`
- `models/deepseek_v4_flash_dspark/decode_compressor_ratio128.py`
- `models/deepseek_v4_flash_dspark/decode_compressor_ratio4.py`
- `models/deepseek_v4_flash_dspark/decode_indexer_compressor.py`
- `models/deepseek_v4_flash_dspark/decode_indexer.py`
- `models/deepseek_v4_flash_dspark/decode_indexer_topk.py`
- `models/deepseek_v4_flash_dspark/decode_sparse_attn_swa.py`
- `models/deepseek_v4_flash_dspark/decode_sparse_attn_hca.py`
- `models/deepseek_v4_flash_dspark/decode_sparse_attn_csa.py`
- `models/deepseek_v4_flash_dspark/decode_swa.py`
- `models/deepseek_v4_flash_dspark/decode_hca.py`
- `models/deepseek_v4_flash_dspark/decode_csa.py`

Standalone fixtures, golden functions, trace-only CLI options, and historical compatibility aliases are validation scaffolding. They should not remain in the final production attention modules after correctness and performance validation has been transferred to tests.

### Static and host checks

- [ ] All public tensor ABIs have at most five dimensions.
- [ ] No production ABI contains a full 1M RoPE table, dense 1M block table, or fixed max-work tensor per request.
- [ ] Page/range/epoch/write-slot validation rejects stale, missing, aliased, inactive, and out-of-range metadata.
- [ ] Geometry covers 1, 127, 128, 129, 12K, 16K, 32K, 512K, 1M-1, and 1M.
- [ ] Heterogeneous request lengths produce independent SWA windows, HCA shard counts, and CSA leaf counts.
- [ ] Block-size-32 page boundaries and ring rollover are explicitly covered.

### Device correctness

- [ ] SWA standalone passes short history, ring wrap, heterogeneous lengths, and 1M tail.
- [ ] Ratio-128 compressor passes no-event, boundary event, state rollover, page permutation, and 1M tail.
- [ ] HCA sparse attention passes 0/1/2/64-shard cases and mixed lengths.
- [ ] CSA selector covers candidate boundaries around 0/1/511/512/513/2048 and the 128-leaf 1M case.
- [ ] CSA sparse-value attention passes 16K and 1M-tail cases.
- [ ] Integrated SWA, HCA, and CSA outputs match their numerical goldens without relaxing existing thresholds.
- [ ] Inactive requests submit no persistent write and preserve the output sentinel.

### Work and trace evidence

- [ ] SWA task/workspace shape is constant once the 128-row window is full.
- [ ] HCA work scales with the actual number of 128-row compressed shards.
- [ ] CSA work scales with the actual number of 2,048-candidate leaves.
- [ ] Representative traces cover 128, 16K, and 1M logical contexts.
- [ ] No heap/task/dependency/TensorMap overflow or scheduler deadlock occurs in the supported attention cases.

## Non-goals

- Serving end-to-end request scheduling and allocator integration.
- Decode CP/TP/EP communication topology (tracked by P3 in #905).
- Prefill 1M enablement (tracked by P4/P5 in #905).
- MTP/DSpark verifier and rollback integration.
- Separate 128/16K/1M kernel profiles or compatibility with the old dense 16K cache ABI.
- Increasing PyPTO's maximum tensor dimension as a workaround.

## Completion definition

This issue is complete when all three standalone attention paths use block-32 ragged cache metadata, execute the exact work implied by heterogeneous runtime lengths up to 1M, pass their existing numerical thresholds, and have no remaining production dependency on the old dense 16K ABI.

sjduan:
Summary:
- Enable elastic long-context SWA, HCA, and CSA decode attention up to 1M positions.
- Use token-local and event-local RoPE inputs with ragged cache and state metadata.
- Add active exact-forest infrastructure for CSA without materializing full-context score tensors.
- Preserve the baseline-shaped standalone attention entry points and delayed cache commit ordering.

Current validation:
- Python syntax compilation passed for the seven CSA files.
- git diff --check passed.
- Phase L records Serving-B compile, runtime, and output validation passes for CSA at B4/S8 128, B4/S8 16K, and B1/S8 1M, including repeated 16K and 1M runs.
- pre-commit was not run because it is unavailable in the local environment and was explicitly waived for this draft.

Known draft blockers:
- Correct CSA compressor event RoPE to use compression-block start 4r rather than the boundary token 4r+3, then update the golden independently.
- Replace the standalone indexer structural fallback with a strict cutoff-equivalent Top-512 candidate validation when the PyPTO two-output loop-carry limitation is resolved or safely bypassed.
- Rebind final source SHAs to fresh device artifacts and complete cold, boundary, rollover, stale-page, and page-permutation coverage.
- Remove nonessential metadata and historical workaround code to reduce the diff before marking the PR ready for review.

Fixes #962


zhangqi-chen (commented):
## Review: scope is far larger than "enable 1M decode attention"

`+10804 / -3822` across 18 files. A large share of that is not 1M support: dead
code, a ceiling-sized constant table, an orthogonal o_proj fusion, a delayed
cache-commit redesign, and renames. Below are 14 concrete points, then a split
proposal.

All counts below are measured against `5791a8ee` (the PR base).

---

### 1. `config.py` carries 84 new constants, half of them dead

- **42 of the 84** new constants have no reference outside `config.py` — they are
  only read by their own asserts. Excluding the 11 that are used solely by
  `context_geometry.py` (see #2, which should go away), **63% are dead**.
  Examples: `CSA_PERSISTENT_BYTES_WITH_SWA_AT_1M_PER_REQUEST_PER_LAYER`,
  `HCA_MAIN_BYTES_AT_1M_PER_REQUEST_PER_LAYER`, `SWA_KV_BYTES_PER_ROW`,
  `CSA_TOPK_PAIR_BYTES`, `TOPK_READY_FRONTIER_W`.
- **69 import-time `assert` statements**, most of them tautological — `CSA_MERGE_ARITY = 2`
  immediately followed by `assert CSA_MERGE_ARITY == 2`, or
  `assert CSA_PERSISTENT_BYTES_AT_1M_PER_REQUEST_PER_LAYER == 303120384`.
  These restate the line above them and freeze magic numbers into the file.
  Kernel files in this repo do not carry asserts.
- Four helpers (`encode_swa_overlay_source`, `decode_swa_overlay_source`,
  `is_swa_persistent_source`, `is_swa_overlay_source`) are called **only by those
  asserts**. The device path computes the overlay inline from
  `SWA_SOURCE_OVERLAY_BASE` and never calls them.

`config.py` should hold only what is genuinely shared across files. `CSA_*` belongs
in `decode_csa.py`, `HCA_*` in `decode_hca.py`, the top-k forest constants in
`decode_indexer_topk.py`.

### 2. `context_geometry.py` should not exist

1950 lines, 53 functions and ~15 dataclasses. **Exactly two functions are imported
anywhere in the repo** — `admit_ragged_page_counts` and `hetero_length_starts_values`,
both by `utils.py`. The remaining ~1900 lines have zero call sites and no test
coverage (there is no dspark contract test).

Move those two functions into `utils.py` and delete the rest. If the derivation
reference has value, it belongs in `docs/`, not as an unreferenced module in the
model directory.

### 3. The HCA ABI change is disproportionate to what 1M requires

HCA replaces `cmp_sparse_indices [T, CMP_TOPK]` + `cmp_block_table` with nine
ragged packed-work tensors (`query_request_ids`, `hca_pages`, `hca_page_offsets`,
`hca_windows`, `request_epochs`, `hca_query_work_offsets`, `hca_work_query_ids`,
`hca_work_row_begin`, `hca_work_valid_rows`).

But HCA compresses at ratio 128, so **1M yields at most 8192 rows**. A dense
`[T, 8192] INT32` index list is 4 MB — it does not need a ragged packing layer.
The candidate stream that genuinely forces ragged work is ratio-4 CSA (262144
candidates).

Please keep HCA on the dense index path and confine the packed-work ABI to CSA.
`decode_csa.py` also goes from **7 to 20 dynamic axes**, which should shrink once
the forest is simplified (#4).

### 4. The Top-K forest is far more complex than the problem needs

The merge is a binary tree over at most 128 leaves, and each merge is an exact
Top-512 over 1024 pairs. The PR encodes this as **four distinct descriptor tables**
(`leaf` / `pair` / `singleton` / `upper`) plus a root table, three `*_actual_count`
scalars, nine dynamic axes, a hand-rolled width-8 pair-wave grid, and a separate
root materialization step.

Top-512 merging is associative, so two much simpler formulations are exactly
equivalent:

- a level-by-level loop — `for level in pl.range(7): spmd(nodes_at_level)` — with
  one descriptor table; or
- a running accumulator (leaf `i` produces top-512, merge into the accumulator),
  which loses only log-depth parallelism and is fine given that typical lengths
  produce a single-digit leaf count.

Two smaller things in the same file:

- `materialize_topk_root` emulates a comparison with fp32 `div` + `trunc` to do the
  sentinel fixup. This is hard to read; a mask/select would be clearer.
- `CSA_MAX_CANDIDATES_FP32 = 262144.0` with `assert CSA_MAX_CANDIDATES_FP32 == CSA_MAX_CANDIDATES`
  is a hardcoded duplicate of a value already available.

### 5. An o_proj fusion is bundled in, orthogonal to 1M and unmeasured

On base, `sparse_attn_csa` returns `o_packed_heads` and `decode_csa_tp1` calls
`decode_o_proj_tp1` separately. In this PR, o_proj is **fused into**
`sparse_attn_csa` and `sparse_attn_hca` (new `sparse_attn_*_local_o_proj`, with the
attention body split into `sparse_attn_*_heads`); `sparse_attn_swa` is left alone.

This accounts for much of the `+910/-460` in `decode_sparse_attn_hca.py` and
`+695/-255` in `decode_sparse_attn_csa.py`, and it has nothing to do with context
length. There is no performance data anywhere in the PR. A structurally similar
fusion landed and had to be reverted recently for costing 12.5% on decode
(#975 / #978), so this needs its own PR with before/after numbers.

Relatedly, `decode_csa_tp1` / `decode_csa_tp1_test` / `golden_decode_csa_tp1` were
renamed to `attention_csa` / `attention_csa_test` / `golden_attention_csa` while
`decode_swa_tp1` and `decode_hca_tp1` kept their names. Please keep the three
families symmetric.

### 6. Raising `max_position_embeddings` to 1M hits files this PR does not touch

`FLASH.max_position_embeddings` goes `16384 -> 1_048_576`, but there are still
**64 `[MAX_SEQ_LEN, ...]` tensor declarations across 11 files**. `freqs_cos` /
`freqs_sin` go from `[16384, 64]` bf16 (2 MB) to `[1M, 64]` (**128 MB each**, 256 MB
per entry).

The affected files include several the PR never modifies:
`prefill_swa.py`, `prefill_csa.py`, `prefill_compressor_ratio128.py`,
`prefill_hca.py`, `dspark_context_kv.py`, `dspark_attention.py`.

`prefill_compressor_ratio128.py` also derives
`HCA_STATE_MAX_BLOCKS = (MAX_SEQ_LEN + 7) // 8`, which jumps from 2048 to 131072,
and its host fixture loops `for block in range(HCA_STATE_MAX_BLOCKS)`. Note
`decode_metadata.py` hardcodes `HCA_STATE_MAX_BLOCKS = 2048`, so the two are now
inconsistent.

Either convert every entry to token-local RoPE inputs first, or scope the ceiling
bump so it cannot reach the prefill entries. Right now there is no evidence the
prefill entries still build.

### 7. `decode_hca.py` is not converted, and violates the invariant this PR states

`config.py` in this PR states the rule explicitly:

> never materialized as a dense per-B table (no `[B, max_logical_pages]` layout)

but `decode_hca.py` still has:

- `freqs_cos` / `freqs_sin` typed `[MAX_SEQ_LEN, ROPE_HEAD_DIM]` — i.e. `[1M, 64]`,
  despite the summary claiming token-local RoPE inputs
- `COMPRESS_TOPK = MAX_SEQ_LEN // COMPRESS_RATIO`, which goes from 128 to **8192**
- `compress_state_block_table: [B_DYN, 131072] INT32` — 8.4 MB at B=16
- `cmp_block_table: [B_DYN, 32768] INT32` — 2.1 MB

### 8. A 16.7 MB static GM arena, allocated regardless of actual length

`TOPK_ARENA_ROWS = CSA_MAX_QUERIES * CSA_MAX_NODES_PER_QUERY = 4080`, so
`pair_arena: [4080, 1024] FP32` is **16.7 MB** resident for every run, at any
context length. This is the same "ceiling must not appear as a runtime quantity"
rule from #7, applied to workspace instead of tables.

### 9. `CSA_MAX_QUERIES = 16` serializes a 128-token step into 8 chunks

`CSA_INDEXER_CHUNK_T = CSA_MAX_QUERIES = 16` while `T = B * S = 16 * 8 = 128`, so one
decode step becomes 8 sequential chunks, each with a 7-level serial merge inside.
This serialization exists only to fit the arena and task-array bounds, and **the PR
contains no performance data at all** to show what it costs.

The standalone indexer test only runs <= 16 tokens (single chunk) — see the comment
at `decode_indexer.py:1034` — so the multi-chunk path has no standalone coverage.

### 10. Dropping `topk_scores` weakened the validation bar

`topk_scores` and `topk_indices` used to be a paired `pl.Out`. Because of a PyPTO
two-output loop-carry phi bug, `topk_scores` was removed, and the check became a
**~150-line custom comparator** that recomputes reference scores inside the compare
function and **accepts index-set differences** when the reconstructed cutoff scores
are "equivalent within tolerance".

Two problems: the oracle is now nearly as complex as the kernel, and it shares its
projection assumptions with the golden it is supposed to check.

The right move is to file the PyPTO loop-carry bug upstream with a minimal repro and
keep the strict paired-output check, rather than relax the comparison. If the strict
check truly cannot run until that is fixed, that should be stated as a blocking
dependency, not absorbed into the test.

### 11. Internal process vocabulary leaked into the shipped ABI

`PHASE_D_*` is used as **ABI identifier names** — 56 distinct names, 165 references
(`PHASE_D_LEAF_FIELDS`, `PHASE_D_PAIR_LEFT_SLOT`, ...). Together with docstrings
referencing "Run 055 Phase A-D", "Phase B/C/D", "Phase D.1", there are 195
occurrences of internal planning vocabulary in product code.

Please rename to something that describes the data (`CSA_LEAF_*`, `TOPK_*`) and drop
the phase/run references from the docstrings — they are meaningless to anyone
reading the file later.

### 12. The PR's own three blockers are still open

From the description:

- CSA compressor event RoPE uses boundary token `4r+3` instead of compression-block
  start `4r`. **This is a numerical error**, and the golden has to be updated
  independently of it.
- The standalone indexer still uses a structural fallback rather than strict
  cutoff-equivalent Top-512 validation.
- pre-commit was not run.

A PR carrying a known numerical error should not be in review. Please land the RoPE
fix first (it is small and separable) and run pre-commit.

### 13. Rename churn, and two divergent SWA ABIs in one directory

`swa_indices -> swa_sources`, `swa_slot_mapping -> swa_write_slots` and similar
renames inflate the diff without changing behaviour. They also leave
`dspark_attention.py` on the old dense `swa_indices` ABI while `decode_swa.py` moves
to `swa_sources` + negative overlay encoding — two SWA source conventions coexisting
in the same model directory.

### 14. The delayed cache commit is an independent change

Moving the current-step KV write from before attention to after it, plus the negative
overlay source encoding, plus the `task_dummy` fences and `allow_early_resolve`
hints, are correctness/scheduling changes that stand on their own. They are not
required by 1M and deserve their own PR with their own validation and numbers.

---

## Suggested split

| # | Content | Size |
|---|---|---|
| 1 | Fix the compressor event RoPE (`4r+3 -> 4r`) and update the golden | Small — land first |
| 2 | Token-local `rope_cos/sin` across the **whole** directory (prefill + `dspark_attention` included), *then* raise `max_position_embeddings` | Medium |
| 3 | Delayed cache commit + overlay source ABI, with before/after numbers | Medium |
| 4 | CSA 1M forest top-k: single descriptor table + level loop, length-proportional arena, `PHASE_D_*` renamed | Large, but well below the current size |
| 5 | HCA/SWA 1M: show the dense index path is insufficient before introducing ragged packing | Small |
| 6 | o_proj fusion into `sparse_attn_hca` / `sparse_attn_csa`, with before/after numbers | Separate |

Independent of the split, and worth doing regardless:

- delete `context_geometry.py` (keep the two used functions in `utils.py`)
- trim `config.py` and drop the 69 asserts and the four unused helpers
- file the PyPTO two-output loop-carry bug upstream and restore the strict Top-K check
- add performance data — there is currently none, and several changes here
  (chunk serialization, the o_proj fusion, the 16.7 MB arena) are plausible regressions

PR #980, opened 2026-08-18
Perf: fuse DeepSeek V4 CSA O projection reduce-scatter

wangqin1723-max:
- Dequantize O-B tiles directly into each reduce-scatter owner window
  instead of materializing and rereading an fp32 projection buffer
- Publish with 24 vector workers, then reduce source lanes in a fixed
  fp32 rank order before the final bf16 cast
- Expose frozen-golden capture and replay for individual CSA cases

CSA output latency 8832.3 -> 8613.7 us at TP2 and 5178.6 ->
5035.6 us at TP4 (a2a3 max case, 100 rounds, 5 warmups,
fastest-rank mean, frozen golden replay). The measurements used different
auto-selected physical card sets, so the 2-3% delta is directional.

### DIFF
--- a/models/deepseek_v4_flash_dspark/decode_csa.py
+++ b/models/deepseek_v4_flash_dspark/decode_csa.py
@@ -71,8 +71,7 @@ from decode_o_proj import (
     LOCAL_T_PAD,
     O_WINDOW_ROWS,
     attention_token_head_all_to_all_step,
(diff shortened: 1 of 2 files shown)


wangqin1723-max:
Closing as duplicate of PR #998, which contains the superseding implementation.

PR #998, opened 2026-08-20
Perf: fuse DeepSeek V4 decode O projection reduce-scatter

wangqin1723-max:
- Replace the split `decode_o_proj` + `o_proj_reduce_scatter` pair with a
  fused `decode_sharded_o_projection_reduce_scatter`, used by the CSA,
  HCA, and SWA decode entry points.
- Dequantize each O-B tile in one of 24 publish workers, stage it in a
  per-worker GM lane, and put it straight into the peer reduce window,
  removing the full `[GROUP_T_PAD, D]` FP32 projection buffer.
- Notify peers only after the staged puts have drained, then reduce the
  window in 8 workers over 4096-column tiles and cast to BF16.
- Drop `decode_o_proj_test`, `build_o_proj_tensor_specs`, and the
  `o_window` / `o_signal` legs of the attention collectives fixture;
  `decode_o_proj_tp1` and the TP1 path stay as they were.
- Return the cache-write task id from `compressor_ratio4` and fence the
  CSA `qk_pv` stage on it together with the original KV writeback, with
  the sparse attention and o-group all-to-all pulled into the consuming
  runtime scope.
- Build the HCA and SWA TP2/TP4 distributed fixtures from the ambient
  RNG instead of a fixed per-rank seed, so each invocation exercises
  fresh inputs while keeping rank replication and O-weight sharding.
- Raise the CSA `x_out` comparator's `max_diff_hd` from 1 to 2, leaving
  `ratio_reldiff(diff_thd=4e-3, pct_thd=0.008)` unchanged; HCA and SWA
  keep `max_diff_hd=1`.

On real A2/A3 TP4 cards 1,9,11,13 with frozen golden data, 5 warmups and
100 timed rounds, fastest-rank `effective_us` means fall from 3805.8 to
3295.6 for CSA (13.4%), 4548.6 to 3872.6 for HCA (14.9%), and 2821.5 to
2308.5 for SWA (18.2%). Slowest-rank-per-round headline means fall from
8165.9 to 7433.2 for CSA and 4425.6 to 3989.8 for SWA, and rise from
7331.0 to 7575.0 for HCA on start skew; the CSA baseline carries the
same cache-ready fence.

### DIFF
--- a/models/deepseek_v4_flash_dspark/decode_csa.py
+++ b/models/deepseek_v4_flash_dspark/decode_csa.py
@@ -62,10 +62,9 @@ from decode_o_proj import (
     LOCAL_T,
     LOCAL_T_PAD,
     O_WINDOW_ROWS,
-    decode_o_proj,
     decode_o_proj_tp1,
+    decode_sharded_o_projection_reduce_scatter,
     o_group_a2a,
-    o_proj_reduce_scatter,
 )
 from decode_sparse_attn_csa import (
     ROPE_CS_T_TILE,
@@ -262,6 +261,8 @@ def decode_csa(
     kv = pl.create_tensor([t_dim, HEAD_DIM], dtype=pl.BF16)
     qr = pl.create_tensor([t_dim, Q_LORA], dtype=pl.INT8)
     qr_scale = pl.create_tensor([t_dim, 1], dtype=pl.FP32)
+    position_ids_t1 = pl.reshape(position_ids, [t_dim, 1])
+    attention_local_flat = pl.create_tensor([ATTENTION_WINDOW_ROWS, O_GROUP_IN], dtype=pl.BF16)
     with pl.scope():
         late_dep = pl.system.task_dummy(deps=[rope_tid])
         qkv_proj_rope(
@@ -272,7 +273,8 @@ def decode_csa(
 
         ori_block_num = pl.tensor.dim(kv_cache, 0)
         kv_cache_flat = pl.reshape(kv_cache, [ori_block_num * BLOCK_SIZE, HEAD_DIM])
-        for wb_blk in pl.spmd(wb_blocks, name_hint="csa_cache_writeback"):
+        with pl.spmd(wb_blocks, name_hint="csa_cache_writeback") as ori_cache_write_tid:
+            wb_blk = pl.tile.get_block_idx()
             wb_t0 = wb_blk * CSA_WB_TOKEN_TILE
             for write_dt in pl.range(CSA_WB_TOKEN_TILE):
                 write_t = wb_t0 + write_dt
@@ -284,7 +286,7 @@ def decode_csa(
                     ]
 
         cmp_out = pl.create_tensor([t_dim, HEAD_DIM], dtype=pl.FP32)
-        compressor_ratio4(
+        cmp_out, cmp_cache_write_tid = compressor_ratio4(
             x_normed_t, cmp_out,
             compress_state, compress_state_block_table,
             cmp_wkv, cmp_wgate, cmp_ape, cmp_norm_w,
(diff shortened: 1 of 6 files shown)


models/deepseek_v4_flash_dspark/decode_csa.py
```diff
@@ -145,6 +149,157 @@ def _parse_tp_argv():
     raise ValueError(f"CSA token capacity {T_PAD} must equal TP local token capacity {LOCAL_T_PAD}")
 
 
+@pl.jit.inline
+def decode_csa_output(
```
zhangqi-chen: remove this and its test.

models/deepseek_v4_flash_dspark/decode_csa.py
```diff
@@ -1774,27 +2238,44 @@ def build_full_compare(mapping_shape, *, leading_rank_axis, diagnostic_x_out=Fal
...
-            runtime_cfg=dict(platform=args.platform),
-            rtol=1e-2,
-            atol=1e-2,
-            compare_fn=build_full_compare(
-                (TP_SIZE, local_t),
-                leading_rank_axis=True,
-                diagnostic_x_out=args.platform.endswith("sim"),
-            ),
-        )
+        if args.entry == "full":
```
zhangqi-chen: do not change this back to full, keep the default test using tp2.

models/deepseek_v4_flash_dspark/decode_csa.py
```diff
@@ -70,13 +70,17 @@ def _parse_tp_argv():
     LOCAL_T,
     LOCAL_T_PAD,
     O_WINDOW_ROWS,
-    decode_o_proj,
     decode_o_proj_tp1,
+    decode_sharded_o_projection_reduce_scatter,
```
zhangqi-chen, line 66: change hca and swa together, then remove the default decode_o_proj, o_proj_reduce_scatter.

zhangqi-chen (changes requested):

models/deepseek_v4_flash_dspark/decode_csa.py
wangqin1723-max: Fixed in `6527740`: migrated CSA, HCA, and SWA together to the fused projection/ReduceScatter helper, then removed the legacy `decode_o_proj` and `o_proj_reduce_scatter` entry points and their standalone tests.

models/deepseek_v4_flash_dspark/decode_csa.py
wangqin1723-max: Fixed in `6527740`: removed `decode_csa_output` and its standalone test, golden/spec helpers, and CLI validation route.

models/deepseek_v4_flash_dspark/decode_csa.py
wangqin1723-max: Fixed in `6527740`: kept TP2 as the default validation path and left larger TP sizes opt-in through `--tp`.

models/deepseek_v4_flash_dspark/decode_o_proj.py
```diff
@@ -663,6 +672,237 @@ def decode_o_proj(
...
+
+        for owner_rank in pl.range(TP_SIZE):
+            if owner_rank != tp_rank:
+                pld.system.notify(
+                    target=reduce_signal,
+                    peer=group_base + owner_rank,
+                    offsets=[tp_rank, 0],
+                    value=1,
+                    op=pld.NotifyOp.AtomicAdd,
+                )
```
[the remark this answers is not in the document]
wangqin1723-max: Fixed in `6527740`: replaced the non-draining computed-tile `remote_store` with per-worker GM staging plus `pld.tensor.put` (with bounded chunks), so each payload drains before that worker notifies its peers.

zhangqi-chen (approved):

PR #1001, opened 2026-08-21
Add: DeepSeek-V4 dspark context-parallel prefill KV all-gather

MaxwellF1:
## Summary

A DSA-CP prefill layer keeps its queries on the local token slice but builds KV
from the **full** token stream: `wkv`, the compressor and the indexer cache all
read every token, so the local slice has to be gathered first. Per #905 P5 that
gather is the layer's **only** activation collective — the `o_proj` epilogue is
token-split and needs none.

This PR lands that one step plus a standalone fixture for it, so the
communication contract can be validated on its own before any attention program
depends on it. It adds one new file and touches nothing else.

## What it does

- `prefill_kv_token_allgather_step` — same shape as the collective steps in
  `decode_o_proj.py`: put a round, barrier, read back, barrier, reset. The
  barriers themselves are **reused** from `decode_o_proj`
  (`tp_group_barrier` / `reset_tp_group_signal`) rather than restated, as
  `decode_{csa,hca,swa}` already do — the credit protocol is subtle and there
  should be one copy of it.
- **The row count comes from the input's own extent**, not a separate scalar, so
  shape and count cannot disagree; one compiled program serves any prefill chunk
  length.
- **Two dynamic token symbols.** The local slice and the gathered stream carry
  different extents in the same program, so they cannot share one — the same
  split #924 made in `qkv_proj_rope` for the q and kv branches.
- **The capacity bound is `prefill_csa.PREFILL_MAX_TOKENS`.** It belongs to the
  *gathered* stream rather than to the per-die slice, because the gather's
  consumers are the replicated KV operators and those are validated to that many
  rows. Taking it from `prefill_csa` keeps the two coupled instead of letting a
  second constant drift.

### Read-back tiling

The decode step copies one row per iteration, which costs 512 iterations there
but would cost `PREFILL_MAX_TOKENS` here. Whole tiles take the wide path and only
the ragged remainder falls back to single rows, so the tail never drags the
aligned body down with it. `--readback-tile 1` reproduces the decode behaviour,
so the two variants are A/B-able from one source rather than from two branches.

## Validation

Eight token counts per group size, chosen to cover the degenerate case, the
`READBACK_ROW_TILE` boundary, non-powers-of-two spanning several tiles, and
capacity with its tail — `1 / 15 / 17 / 129 / 257 / 1024 / cap-1 / cap`:

| Platform | Group | Result |
|---|---|---|
| **a2a3** | TP2 | **8/8 PASS** |
| **a2a3** | TP4 | **8/8 PASS** |
| a2a3sim | TP2 | 8/8 PASS |
| a2a3sim | TP4 | 8/8 PASS |

`ruff check --config ruff.toml`, `tests/lint/check_headers.py` and
`tests/lint/check_english_only.py` are all clean.

The fixture's per-rank rows carry their rank in the leading column, so a gather
that lands a rank's block at the wrong offset fails the comparison instead of
averaging out.

## Performance

a2a3, even cards, TP2, `group_t = 8192`, 100 rounds / 5 warmup, **fastest rank**,
baseline `--readback-tile 1`:

| `READBACK_ROW_TILE` | fastest-rank mean | vs tile 1 |
|---|---|---|
| 1 (decode's row-at-a-time) | 5946.8 µs | — |
| 8 | 3660.4 µs | 1.62× |
| **16** (default) | **3023.2 µs** | **1.97×** |

Tile 32 and above exceed the Vec buffer limit at this width
(`262144 > 188416 bytes`), so 16 is both the default and the ceiling here.

## Notes

- The step has no in-tree consumer yet; the SWA / HCA / CSA context-parallel
  paths that call it are follow-ups. `decode_o_proj.py`'s own
  `kv_token_allgather_step` sits in the same state today.
- `ci: devices=2` matches `decode_csa.py` / `moe.py` / `lm_head.py`; the fixture
  also runs at `--tp 4`.

Part of #905 (P5).

### DIFF
--- a/models/deepseek_v4_flash_dspark/prefill_kv_allgather.py
+++ b/models/deepseek_v4_flash_dspark/prefill_kv_allgather.py
@@ -1,0 +1,270 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# -----------------------------------------------------------------------------------------------------------
+# ci: devices=2  # CI: 2-card run; borrows 2 cards via task-submit --device-num
+"""Context-parallel prefill KV-stream all-gather.
+
+A DSA-CP prefill layer keeps its queries on the local token slice but builds KV
+from the *full* token stream: `wkv`, the compressor and the indexer cache all
+read every token, so the local slice has to be gathered first. That gather is
+the layer's only activation collective -- the o_proj epilogue is token-split and
+needs none.
+
+This module carries that one step plus a standalone fixture for it, so the
+communication contract can be validated on its own before any attention program
+depends on it.
+
+Relative to the decode-side `kv_token_allgather_step` (`decode_o_proj.py`) the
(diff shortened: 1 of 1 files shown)


MaxwellF1:
Closing: opened prematurely, before the author had reviewed the change. Will re-open after review.

Issue #1004, opened 2026-08-21
[Feature] DeepSeek-V4 dspark prefill context parallelism (#905 P5): DSA-CP across SWA / HCA / CSA


MaxwellF1:
## Summary

Implement #905 **P5 — prefill parallelism and communication** for
`models/deepseek_v4_flash_dspark/`: bring the three prefill attention blocks
(SWA / HCA / CSA) under DSA context parallelism, so a prefill step's query work
divides across the 4-die TP group while KV stays replicated.

The requirement lives in #905 P5. This issue tracks, requirement by requirement,
how far the directory satisfies it, and links the PRs that get it there.

## Contract

P5 leaves one decision to this item: *"Deciding which contract this directory
implements is part of this item."*

**Decision: replicated KV**, matching vllm-ascend's
`attention/context_parallel/dsa_cp.py`.

| | Replicated KV (chosen) | Exchanged KV (`deepseek_v4_flash_mtp/prefill_cp_*`) |
|---|---|---|
| KV cache per die | full (one KV head, 512 dim) | 1/N |
| Cross-die dependency | none | ring, causally ordered segments |
| Activation collectives per layer | 1 all-gather | one KV exchange per segment |
| Redundant compute | `wkv` + compressor + indexer cache on the full stream | none |
| Matches the deployment | yes — `wkv` is `ReplicatedLinear`, `num_kv_heads = 1` | no |

Exchanged KV only pays off for KV-cache memory at 1M context, which is P2.

## Requirement checklist

Each item is a requirement stated in #905 P5, not a work breakdown.

### Layout — P5's per-component table

- [ ] `wq_a`, `wq_b` — CP, split on token
- [ ] attention QK / PV — CP, split on query
- [ ] `wkv`, compressor, indexer cache — no split, full stream, **preceded by an
      all-gather**
- [ ] `wo_a` — CP, split on **token** (not head/o_group as in decode)
- [ ] `wo_b` — CP, split on **token** (not K as in decode)

### Communication

- [ ] **Exactly one activation collective per layer.** *"Prefill has one
      activation collective, decode has three."*
- [ ] **No extra causal communication.** The cross-rank causal dependency is
      paid in redundant compute, not collectives — every die computes the KV it
      needs from the gathered stream.
- [ ] Slot mappings are never sliced to the local range; the replicated segment
      runs at global positions.

### `o_proj` strategy inversion

- [ ] Split axis moves from head/o_group to token, `n_group = 8`
[...]

PR #1007, opened 2026-08-21
Add: DeepSeek-V4 dspark context-parallel prefill SWA

MaxwellF1:
## Summary

Bring the SWA prefill block — the layers with `compress_ratio = 0`, sliding
window only, no compressor — under DSA context parallelism, per #905 P5.

`prefill_swa.py` now routes by `--tp`, the shape `decode_swa.py` already uses:
`--tp 1` runs the single-die tree, `--tp 2` / `--tp 4` the context-parallel one.
No new file.

## What it does

Queries stay on the local token slice; KV is built from the gathered stream.
**Nothing is re-implemented** — every stage is the same inline the single-die
path calls. What differs is the token extent each stage runs at, and the one
all-gather between them:

- **query side, local** — `hc_pre`, the Q projection, the window indices,
  attention, `o_proj`, `hc_post`;
- **KV side, gathered** — `wkv` and the cache write, at global positions through
  an unsliced slot mapping.

The two views need separate dynamic token symbols because they carry different
extents in one program. `q_proj_rope` and `kv_proj_rope` already take separate
axes since #924, so the split needs no change there.

Feeding the KV side a local slice or local positions would not fail loudly — it
would write the wrong cache rows — so the golden compares `kv_cache` as well as
the layer output.

`o_proj` needs no work: a standalone prefill program already holds the full
weight and emits no all-to-all or reduce-scatter, so P5's inverted `o_proj`
strategy holds by construction.

## Why two trees rather than one

The single-die tree is kept, not replaced, because at `--tp 1` the CP body would
cost a self-addressed all-gather and would lose the fused `qkv_proj_rope`, whose
q and kv branches share one tile loop and one rope table.

A compile-time `if TP_SIZE == 1` inside the body does not avoid that:
`@pl.jit.inline` rewrites Python `if` into a device-side conditional, so both
arms are type-checked and `--tp 1` fails on `Cannot reassign 'kv_full' with a
different type: was [q_dim, 512], got [kv_dim, 512]`. `decode_swa.py` carries two
trees for the same reason.

The diff reflects this: **no line of the existing kernel body, golden or specs is
deleted**. The 27 deleted lines are the module docstring, the import block, the
dynamic-symbol header and the old `__main__`.

## Validation

The golden is the existing single-die reference run over the whole stream and
sliced per rank, not a second implementation. Replicated KV makes the two
**element-wise equal**: each die sees exactly the KV a single die would, and
attention is row-independent with the reduction along K.

a2a3:

| Mode | Tokens | Result |
|---|---|---|
| `--tp 1` | 512 | PASS |
| `--tp 2` | 128, 512 | PASS |
| `--tp 4` | 128, 512 | PASS |

The `--tp 1` case is the regression that shows the single-die tree still behaves
as it did.

`ruff check --config ruff.toml`, `check_headers.py` and `check_english_only.py`
are clean.

## Note on the default

`_TP_DEFAULT` is 2, inherited from `prefill_kv_allgather.py` and matching
`decode_swa.py`, so a bare `python prefill_swa.py -p a2a3` now runs the
context-parallel tree on two cards. `# ci: devices=2` follows from that.

Part of #1004, #905 (P5).

### DIFF
--- a/models/deepseek_v4_flash_dspark/prefill_swa.py
+++ b/models/deepseek_v4_flash_dspark/prefill_swa.py
@@ -6,9 +6,11 @@
 # INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
 # See LICENSE in the root of the software repository for the full text of the License.
 # -----------------------------------------------------------------------------------------------------------
-"""DeepSeek-V4 packed prefill SWA attention over one contiguous run of <=T tokens."""
+# ci: devices=2  # CI: 2-card run; borrows 2 cards via task-submit --device-num
+"""DeepSeek-V4 packed prefill SWA attention: single-die at --tp 1, context-parallel above it."""
 
 import pypto.language as pl
+import pypto.language.distributed as pld
 
 from config import (
     BLOCK_SIZE,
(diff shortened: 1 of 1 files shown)


zhangqi-chen:
Landed in #1010, which squash-merged this branch together with #1009 and the CSA work stacked on top of it. `prefill_swa.py` on `main` (c7f700c4) is identical to this branch's version, so there is nothing left to merge here.

PR #1009, opened 2026-08-21
Add: DeepSeek-V4 dspark context-parallel prefill HCA

MaxwellF1:
## Summary

Bring the HCA prefill block — the layers with `compress_ratio = 128` — under DSA
context parallelism, per #905 P5. `prefill_hca.py` routes by `--tp` exactly as
the SWA companion does.

## What it does

The increment over SWA is one component, and it lands on the KV side:

- the **ratio-128 compressor** consumes the gathered stream at the original
  positions and scatters through the unsliced compressed and state slot
  mappings, so every rank rebuilds the same compressed cache.

The compressed **visibility** per query stays query-side: `visible_cmp` follows
each local query's own position, not the global stream. Those two are easy to
confuse and neither fails loudly when confused — a local slice or local
positions would quietly compress the wrong windows.

So the golden compares `cmp_kv` and `compress_state` as well as the layer
output; a rank that compressed the wrong window shows up there rather than being
averaged out of the attention result.

As with SWA, no stage is re-implemented, the golden is the single-die path run
over the whole stream and sliced per rank, and no line of the existing kernel
body, golden or specs is deleted.

## Validation

a2a3, four outputs compared (`x_out_local`, `kv_cache`, `cmp_kv`,
`compress_state`):

| Mode | Tokens | Result |
|---|---|---|
| `--tp 1` | 512 | PASS |
| `--tp 2` | 128, 512 | PASS |
| `--tp 4` | 128, 512 | PASS |

Lint clean.

This branch is stacked on #1007, so the diff also carries `prefill_swa.py`,
which is under review there. The file this PR changes is `prefill_hca.py`.

Part of #1004, #905 (P5).

### DIFF
--- a/models/deepseek_v4_flash_dspark/prefill_hca.py
+++ b/models/deepseek_v4_flash_dspark/prefill_hca.py
@@ -6,6 +6,7 @@
 # INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
 # See LICENSE in the root of the software repository for the full text of the License.
 # -----------------------------------------------------------------------------------------------------------
+# ci: devices=2  # CI: 2-card run; borrows 2 cards via task-submit --device-num
 """DeepSeek-V4 packed prefill HCA (ratio-128) attention over one contiguous run of <=T tokens."""
 
 import functools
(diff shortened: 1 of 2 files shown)


zhangqi-chen:
Landed in #1010, which squash-merged this branch together with #1007 and the CSA work stacked on top of it. `prefill_hca.py` on `main` (c7f700c4) is identical to this branch's version, so there is nothing left to merge here.

PR #1013, opened 2026-08-22
Add: 1M context support to DeepSeek V4 CSA decode

sjduan:
- Raise the CSA decode context ceiling to 1M positions across the
  compressor, indexer, and sparse-attention paths, deriving every paged
  block-table width from that ceiling instead of the config-level
  KV_CMP_/KV_ORI_/IDX_CACHE_ constants
- Size the main and inner recurrent state as a fixed 8-slot ring rather
  than one slot per position, and size the compressed and index cache
  pools from the requested start positions
- Rewrite the ratio-4 compressor to pool each boundary against the old
  ring plus the current-step projection overlay and commit the ring in a
  separate scope, so a later token in the same step cannot overwrite
  rows an earlier boundary still reads
- Guard cold-start compressor reads so an unmapped state page keeps its
  zero/-inf defaults and a negative cache slot writes a zero KV row
- Replace the full [MAX_SEQ_LEN, ROPE_HEAD_DIM] RoPE table parameters
  with token_local_rope in utils.py, which materializes only the rows
  the supplied absolute positions use
- Select an exact Top-512 over up to 262,144 visible candidates with a
  128-leaf, 255-node mrgsort forest seven merge levels deep, replacing
  the fixed 4096-value two-half selector
- Fan leaf scoring across eight work units per token and fence the AIC
  matmul against the AIV reduction with FFTS events
- Feed the compact [T, IDX_TOPK] result straight into sparse attention
  instead of the [T, MAX_SEQ_LEN // 4] score-length tensor
- Accept per-request start positions in resolve_start_positions as an
  int, list, tuple, or tensor, and validate the entry count against the
  batch
- Add --start-pos to the standalone sparse-attention harness, drop its
  multiple-of-4 batch restriction, and hoist the shared SWA indices and
  window block table out of the fixture initializers

One canonical fixture now serves TP1, TP2, and TP4; the HC, QKV, sparse
QK/PV, output-projection, and TP communication topology are unchanged.
Refs #962.


sjduan:
CI and AI-review follow-up is in commit 434b566.

Related fixes:
- guard unmapped recurrent-state pages in both ratio-4 compressors and align the indexer golden
- keep the default standalone compressor fixtures on a complete recurrent window, avoiding the known position-3 nested pooling lowering edge while explicit start positions remain available
- document the semantic indexer check and allow the observed 0.1 percent diagnostic score budget; the selected Top-K indices passed
- add the static RoPE tile contract, probe-capacity ValueError, CLI start-position validation, and reuse sparse fixture metadata
- remove the unused distributed-path task-id lint finding

Not applied after validation:
- zero-writing negative cache slots, adding the zero-candidate leaf probe, and physical-page guards each caused AICPU failures in otherwise valid device cases on the current PyPTO runtime
- moving shared helpers and changing the 1M ceiling were intentionally skipped to keep the baseline diff minimal and preserve the feature scope

Validation completed: py_compile and diff checks pass; both compressor default B=16 device cases pass; TP2 CSA compile-only passes; CLI and capacity fail-closed checks pass. The current PyPTO HEAD cbcedd3b has a separate AICPU 507018 runtime regression that also reproduces with the original indexer and forest code, so the pinned PR CI rerun is the authoritative device gate.

PR #1023, opened 2026-08-24
Add: DeepSeek-V4 Flash serving contract

vegetabledoww:
- Add `models/deepseek_v4_flash_mtp/serving_contract.py`, a frozen
  dataclass declaring the kernel's prefill limits: a 128-token internal
  tile, at most 8192 active tokens per request per dispatch, one prefill
  request per rank partition, and no prefill/decode mixing within one
  dispatch.
- Validate the manifest at construction: schema version 1, a positive
  tile, and a per-request cap that is a positive multiple of that tile.
- Add `padded_prefill_tokens()`, rounding an active extent up to the tile
  and rejecting non-positive, over-cap, and non-integer inputs including
  `bool`, `float`, and `str`.
- Restate the prefill shape in `docs/models/deepseek_v4_flash_mtp.md` as
  one request per rank partition of up to 8192 active tokens, walked in
  128-token tiles.
- Run the new contract tests alongside the existing golden and
  DeepSeek-V4 Pro contract tests in CI.

The 8192 cap bounds a single prefill dispatch for one request, not the
model's context length; splitting a longer prompt across dispatches stays
with the serving side (hw-native-sys/pypto-serving#187). The module has
no import side effects and changes no kernel compute.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,9 @@ jobs:
           pip install torch --index-url https://download.pytorch.org/whl/cpu
 
       - name: Run golden and DeepSeek contract tests
-        run: pytest tests/golden tests/contract/test_deepseek_v4_pro_*.py -v
+        run: >-
+          pytest tests/golden tests/contract/test_deepseek_v4_pro_*.py
+          tests/contract/test_deepseek_v4_flash_serving_contract.py -v
 
   detect-changes:
     runs-on: ubuntu-latest
--- a/docs/models/deepseek_v4_flash_mtp.md
+++ b/docs/models/deepseek_v4_flash_mtp.md
@@ -16,7 +16,7 @@ that every kernel imports as a bare sibling module.
 | Speculative decoding | MTP = 1 — one draft token verified against the previous one, so a decode step carries `S = 2` token rows per request |
 | Decode batch per card | 4 requests → 8 token rows per step (`DECODE_BATCH`, `DECODE_SEQ`) |
 | Decode context length | up to 16,384 positions, paged in 128-token pages (`max_position_embeddings`, `BLOCK_SIZE`) |
-| Prefill shape | one request of 128 tokens per program (`PREFILL_BATCH`, `PREFILL_SEQ`) |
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -450,13 +452,17 @@ jobs:
         with:
           pypto-lib-checkout-path: checkout
           pypto-serving-checkout-path: serving-checkout
+          # PR #1023 and pypto-serving #187 form one cross-repository contract.
+          pypto-serving-ref: refs/pull/187/head
```
zhangqi-chen: Remove the serving commit bindings

.github/workflows/ci.yml
```diff
@@ -450,13 +452,17 @@ jobs:
...
           # second 8-card server for the same gate.
           deepseek-mtp-cases: k1
-          pto2-ring-dep-pool: "131072"
-          pto2-ring-task-window: "131072"
-          pto2-ring-heap: "2147483648"
+          pto2-ring-dep-pool: "16384,16384,16384,16384"
+          pto2-ring-task-window: "16384,16384,16384,16384"
+          # Keep each dispatch depth below the measured 1 GiB bound instead of
+          # broadcasting one 2 GiB scalar into all four static arenas.
+          pto2-ring-heap: "1073741824,1073741824,1073741824,1073741824"
```
zhangqi-chen: why change it to 1G, or is it useless?

zhangqi-chen (changes requested):

.github/workflows/ci.yml
vegetabledoww: 已移除对 Serving PR #187 的固定 ref，同时撤回了 composite action 中新增的 pypto-serving-ref 参数。Lib 门禁恢复验证 Serving main，跨仓配套验证由 Serving #187 自身门禁负责。

.github/workflows/ci.yml
vegetabledoww: 此前改为每层 1 GiB 是为规避 scalar ring 配置广播后挤占 HBM，但该调整与本 PR 的 contract 职责无关。现已撤回全部 ring 参数修改，恢复既有 CI 基线；如后续仍需调整，会作为独立 CI 修复提交。

zhangqi-chen (approved):

PR #1024, opened 2026-08-24
Refactor: defer DeepSeek V4 MTP communication waits

MaxwellF1:
- Replace the isolated blocking waits for MoE payload dispatch, MoE
  combine, LM-head hidden dispatch, and LM-head logits combine with
  deferred completion.

- Keep each deferred wait in a dedicated top-level CORE_GROUP task. The
  task registers its remote counter conditions, releases the AIV core,
  and remains logically incomplete until every condition is satisfied.
  Existing downstream dependencies continue to use the waiter TaskId.

- Keep the MoE metadata wait blocking because the same task resumes
  after the wait to read the remote counts and build the local cumsum.
  This continuation is incompatible with deferred completion.

- Fold MoE combine completion publication into each of the N_LOCAL
  scatter blocks after its remote puts. Every block atomically publishes
  one completion to each peer, so the monotonic wait threshold becomes
  moe_epoch * N_LOCAL.

- Remove the push/combine TaskId returns, anchor reads, and explicit
  dependencies that were needed only to keep spinning waiters behind
  their matching producers.

- Preserve the existing point-to-point put, notify, and monotonic epoch
  protocol while removing the need to manually order locally independent
  send and wait tasks.

### DIFF
--- a/models/deepseek_v4_flash_mtp/lm_head.py
+++ b/models/deepseek_v4_flash_mtp/lm_head.py
@@ -162,14 +162,10 @@ def lm_head(
                     op=pld.NotifyOp.AtomicAdd,
                 )
 
-    # Barrier on the group's publishes. The hidden_states read is an anchor, not
-    # data: it deps this task on the hidden-state producer so the wait runs
-    # alongside our own push instead of trailing it.
     with pl.at(level=pl.Level.CORE_GROUP, name_hint="lm_head_dispatch_wait") as _dwait_tid:
-        _hidden_anchor = pl.read(hidden_states, [0, 0])
         for owner_tp in pl.range(TP_SIZE):
             if owner_tp != tp_rank:
-                pld.system.wait(
+                pld.system.defer_wait(
                     signal=hidden_done,
                     offsets=[owner_tp, 0],
                     expected=pl.cast(done_epoch * MAX_LOGIT_ROWS, pl.INT32),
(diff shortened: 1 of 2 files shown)


MaxwellF1:
## Validation

The final implementation passed the following MTP golden validations
without a timeout or deadlock:

| Topology | Path | Validated outputs | Result |
|---|---|---|---|
| EP4 | Prefill | kv_cache, hidden_out, pre_hc_hidden_out, logits | PASS |
| EP4 | Decode | kv_cache, hidden_out, next_pre_hc_hidden, logits, sampled_ids | PASS |
| EP8 / TP4 / DP2 | Prefill, 128 tokens | kv_cache, hidden_out, pre_hc_hidden_out, logits | PASS |
| EP8 / TP4 / DP2 | Decode, 8 tokens | kv_cache, hidden_out, next_pre_hc_hidden, logits, sampled_ids | PASS |

The rebased EP8 MoE entry also passed compile-only validation on the
current `upstream/main`.

## EP8 performance reference

The earlier EP8 decode comparison produced the following mean latency:

| Implementation | Mean | Difference |
|---|---:|---:|
| Blocking wait + manual deps | 55.242 ms | baseline |
| Deferred wait | 55.048 ms | -0.35% |
| Soft syncall | 56.027 ms | +1.42% |

The deferred-wait result does not show a meaningful performance change;
the 0.35% difference is within normal run-to-run noise. The purpose of
this change is therefore not a claimed speedup. It removes the need to
manually analyze and maintain dependencies between locally independent
send and wait tasks while preventing waiters from occupying the AIV
cores needed by their matching sends.

Deferred wait was selected instead of soft syncall because it preserves
the existing point-to-point put/notify/counter protocol and changes only
how an isolated waiter completes. Soft syncall introduces a local task
join and changes where synchronization is enforced, so it is a broader
scheduling change rather than a primitive-only replacement.

These EP8 numbers were collected on the earlier deferred-wait candidate.
The final cleanup additionally removes the remaining combine TaskId and
anchor plumbing. The final code has been revalidated for EP8 correctness,
but this table should be treated as a performance reference rather than
a strict final-code A/B.

PR #1025, opened 2026-08-24
Perf: fuse DeepSeek V4 SWA and HCA O projection reduce-scatter

wangqin1723-max:
- Publish dequantized O-B tiles directly into ReduceScatter owner windows.
- Reuse the shared fused O projection and ReduceScatter path for SWA and HCA tensor-parallel outputs.
- Remove the intermediate FP32 projection buffer and separate projection/reduction boundary from both paths.
- Preserve TP1 paths and HCA compressed-cache dependencies.

SWA TP4 a2a3 max on fixed cards 0,1,2,3 with 20 rounds and 5 warmups: headline mean 4197.8 -> 3559.5 us (15.2% lower), fastest-rank mean 2912.0 -> 2084.9 us (28.4% lower), with kv_cache and x_out validation passing.

HCA TP4 max and subcapacity fresh-golden validation passed on fixed cards 0,1,2,3. compress_state, kv_cache, cmp_kv, and x_out all passed in both cases.

### DIFF
--- a/models/deepseek_v4_flash_dspark/decode_hca.py
+++ b/models/deepseek_v4_flash_dspark/decode_hca.py
@@ -66,10 +66,9 @@ from decode_o_proj import (
     LOCAL_T,
     LOCAL_T_PAD,
     O_WINDOW_ROWS,
-    decode_o_proj,
+    decode_sharded_o_projection_reduce_scatter,
     decode_o_proj_tp1,
     o_group_a2a,
-    o_proj_reduce_scatter,
 )
 from decode_sparse_attn_hca import (
     CMP_TOPK,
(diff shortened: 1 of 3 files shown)


wangqin1723-max:
Closing after re-evaluating the remaining idea on current main (3ac8fd88). The broader fused O-projection ReduceScatter design has already landed and evolved through #998 and #1070. Porting the remaining direct remote-store publication path onto the current owner-parallel implementation passed TP4 SWA correctness, but regressed A2/A3 TP4 fastest-rank median latency from 645.3 us to 703.2 us (+9.0%) on devices 1,3,5,9 with 100 timed rounds / 5 warmups and the same frozen golden. The headline median also moved from 1459.0 us to 1503.6 us (+3.1%). The experiment was reverted, so this older conflicting PR is no longer worth carrying.

PR #1035, opened 2026-08-25
Perf: eliminate HCA raw KV GM transpose

wangqin1723-max:
- Load raw KV tiles into Mat memory and use a transpose view for QK
  while reusing the original tile for PV
- Remove the raw_kv_t GM buffer and standalone transpose task
- Resolve the causal-regression sentinel through the window block table

TP4 A2/A3 on devices 3/5/7/9, global batch 64, sequence 8, local
T=128, frozen golden, and 100 rounds after 5 warmups: fastest-rank
median 3906.2 -> 3453.9 us (-11.6%).


wangqin1723-max:
Rebased the focused optimization onto current main (c599d13) and remeasured with the wq3 pin chain.

The TP4 full-HCA replay passed against the same frozen golden. Fastest-rank median improved 3906.2 -> 3453.9 us (-11.6%). The optimized 100-round run also contained synchronized 12-21 ms scheduling outliers, so fastest-rank mean was distorted from 3942.6 to 4929.5 us. A separate level-4 profile selected a 3682.5 us fast rank and found a 1125.6 us ready-to-dispatch delay before hca_gather_kv; no hca_raw_transpose task remains in the optimized graph.

PR #1034, opened 2026-08-25
Perf: fuse attention output with AllToAll

wangqin1723-max:
- Publish HCA output directly from 48 packing workers
- Fold CSA and SWA merge, normalization, inverse RoPE, and packing
  into distributed publishers
- Gather published output with a reusable notification handshake before
  O projection
- Add HCA and CSA frozen-golden CLI forwarding for repeatable timing
- Seed decode runtime buffers explicitly on device and pass ring sizing
  through RunConfig for the current Simpler API
- Unify the CSA, HCA, and SWA sparse-attention helper names around a
  shared intermediates/publish split

Fastest-rank effective mean on a2a3 TP2, local_t=256, devices 1,3,
100 rounds, 5 warmup, a 1 GiB ring, and one frozen golden per path:
HCA 6375.8 -> 4785.9 us (24.9%), CSA 7069.5 -> 4775.9 us
(32.4%), and SWA 2919.4 -> 1797.7 us (38.4%). The CSA baseline
contained 25.95 and 37.44 ms raw outliers; its fastest-rank median
still moved 6412.0 -> 4755.6 us (25.8%).

Fastest-rank effective mean on a2a3 TP4 with frozen golden data:
HCA 3879.8 -> 3012.5 us (20 rounds, 5 warmup; devices 3,5,7,9).
CSA 3295.6 -> 2642.7 us (100 rounds, 5 warmup;
devices 1,9,11,13).
SWA 2308.5 -> 1180.7 us (100 rounds, 5 warmup;
devices 1,9,11,13).

PR #1040, opened 2026-08-25
Perf: pipeline HCA AllToAll into O-A

wangqin1723-max:
- Publish HCA output from packing workers and fold CSA/SWA merge,
  normalization, inverse RoPE, and packing into distributed publishers
- Gather CSA/SWA output with reusable notification handshakes before the
  sharded O projection
- Publish HCA O-A tile readiness and start source-sized O-A launches
  directly from distributed attention windows
- Release HCA A2A credits only after publisher and O-A consumers complete
- Preserve CSA/SWA O-projection ordering while sharing the sharded O-B
  publish and ReduceScatter tail
- Add frozen-golden CLI forwarding for repeatable HCA and CSA timing

TP2 HCA decode median 6434.6 -> 5910.6 us (A2A3 devices 13,15,
frozen golden, 100 rounds, 5 warmup, median of four run medians).
The fused publisher path measured TP4 HCA 3879.8 -> 3012.5 us
(20 rounds, 5 warmup; devices 3,5,7,9), CSA 3295.6 -> 2642.7 us,
and SWA 2308.5 -> 1180.7 us (100 rounds, 5 warmup;
devices 1,9,11,13).

### DIFF
--- a/models/deepseek_v4_flash_dspark/decode_csa.py
+++ b/models/deepseek_v4_flash_dspark/decode_csa.py
@@ -64,12 +64,13 @@ from decode_o_proj import (
     O_WINDOW_ROWS,
     decode_o_proj_tp1,
     decode_sharded_o_projection_reduce_scatter,
-    o_group_a2a,
+    o_group_a2a_finish,
 )
 from decode_sparse_attn_csa import (
     ROPE_CS_T_TILE,
     T_PAD,
     sparse_attn_csa,
+    sparse_attn_csa_a2a,
 )
 
 # Dynamic shape variables.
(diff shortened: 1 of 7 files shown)


wangqin1723-max:
This PR is stacked on #1034. The receive-side HCA pipeline is the top
commit, c1f79b2; the branch was rebased onto main at dfd3a87 without
modifying the existing #1034 branch.

TP2 A2A3 capacity validation used the same devices (13,15), frozen golden,
100 rounds, and 5 warmup rounds for every baseline/variant run. The two
counterbalanced orders produced these effective-time medians:

- B-V-V-B: 8671.1, 6148.8, 5508.2, 6382.7 us
- V-B-B-V: 5672.3, 6486.4, 6194.8, 6495.2 us

The median of the four baseline run medians is 6434.6 us; the median of the
four fused run medians is 5910.6 us (-8.1%). The raw runs show substantial
host/device phase noise, so the individual medians are included rather than
reporting only the aggregate.

Level-4 traces confirm physical publisher/O-A overlap of 79.52 us on rank 0
and 170.60 us on rank 1, versus 0 us in #1034. The profiled run itself had
large rank skew and is used only as topology evidence, not as the performance
number.

Real-device correctness passed for TP2 capacity, sub-capacity (local_t=248,
100 persistent rounds), and minimum shape (local_t=8). A2A3 compile-only
passed for TP2 and TP4 capacity/sub-capacity. Simulator runs were intentionally
left out of this validation scope.

PR #1055, opened 2026-08-26
Add: integrate DeepSeek V4 D-Spark full decode forward

sjduan:
- add models/deepseek_v4_flash_dspark/decode_fwd.py: the full D-Spark
  decode forward over all 43 layers in the fixed order SWA, SWA,
  (CSA, HCA) x 20, CSA, reusing the production SWA/CSA/HCA layer entries
  without duplicating attention, compressor, indexer, output-projection,
  or MoE math
- run one rank-local L2 per rank with packed per-layer cache and state
  pools, activation ping-pong, and MoE epochs 1 through 43; the layer
  ordinal is packed into the block extent so public tensor ABIs stay at
  five dimensions or fewer
- keep model-layer and attention-kind ordinals separate, so layer 42
  maps to CSA ordinal 20
- allocate the attention/O-projection, MoE, and LM-head windows once at
  L3 and reuse them with independent epoch domains; each rank publishes
  one full-forward child rather than one child per layer
- connect the terminal HC head, RMSNorm, TP LM head, greedy sampling,
  inactive-row masking, and the single final MoE signal clear
- select the topology with --tp, --ep, and --weight-bank-size (1 for the
  shared-bank fixture, 43 for the production per-layer banks), and mark
  the entry device-only at EP2/TP2 with ci: devices=2 and ci: no-sim
- lm_head: accept TP1 alongside the existing TP2/TP4 paths, and pass the
  1 GiB ring heap through the runner's ring_heap argument instead of
  defaulting PTO2_RING_HEAP in the module
- hc_head: seed the split-K accumulator with an explicit spmd zero-fill
  that the projection tasks depend on, replacing the create_tensor
  init_value
- expert_shared: drop the redundant init_value on the requantized hidden
  tile

Related to #962.

### DIFF
--- a/models/deepseek_v4_flash_dspark/decode_fwd.py
+++ b/models/deepseek_v4_flash_dspark/decode_fwd.py
@@ -7,6 +7,7 @@
 # See LICENSE in the root of the software repository for the full text of the License.
 # -----------------------------------------------------------------------------------------------------------
 # ci: devices=2  # CI: EP2/TP2 full decode forward
+# ci: no-sim    # CI marker: full multi-layer / multi-card forward — device-only, skip on *sim
 """DeepSeek-V4 D-Spark decode forward."""
 
 import sys
--- a/models/deepseek_v4_flash_dspark/decode_fwd.py
+++ b/models/deepseek_v4_flash_dspark/decode_fwd.py
@@ -34,12 +34,18 @@ def _parse_parallel_arg(name, default):
 # importing attention, MoE, or output-projection modules.
 TP_SIZE = _parse_parallel_arg("tp", _TP_DEFAULT)
 EP_SIZE = _parse_parallel_arg("ep", _EP_DEFAULT)
+FWD_WEIGHT_BANK_SIZE = _parse_parallel_arg("weight-bank-size", 1)
 if TP_SIZE not in _TP_CHOICES:
     raise ValueError(f"--tp must be one of {_TP_CHOICES} (got {TP_SIZE})")
 if EP_SIZE not in _EP_CHOICES:
     raise ValueError(f"--ep must be one of {_EP_CHOICES} (got {EP_SIZE})")
 if EP_SIZE % TP_SIZE != 0:
     raise ValueError(f"EP={EP_SIZE} must be divisible by TP={TP_SIZE}")
+if FWD_WEIGHT_BANK_SIZE not in (1, 43):
+    raise ValueError("--weight-bank-size must be 1 or 43")
+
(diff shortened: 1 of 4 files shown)


models/deepseek_v4_flash_dspark/decode_fwd.py
```diff
@@ -0,0 +1,3348 @@
...
+    "hca_cmp_kv": 2,
+    "csa_compress_state": 4,
+    "csa_cmp_kv": 2,
+    "csa_inner_compress_state": 4,
+    "csa_idx_kv_cache": 1,
+    "csa_idx_kv_scale": 4,
+}
+
+
+def build_packed_pool_layout(per_layer_shapes):
```
zhangqi-chen: remove all unused/debugging functions

models/deepseek_v4_flash_dspark/decode_fwd.py
```diff
@@ -0,0 +1,3348 @@
...
+            routed_w2,
+            [N_LOCAL, D, MOE_INTER],
+            [weight_layer_swa0 * N_LOCAL, 0, 0],
+        )
+        raw_kv_layer_swa0 = pl.slice(
+            raw_kv_pool,
+            [raw_blocks_per_layer, BLOCK_SIZE, 1, HEAD_DIM],
+            [0, 0, 0, 0],
+        )
+        decode_layer_swa(
```
zhangqi-chen: do not import decode layer, decode-layer.py is for debugging precision only.

models/deepseek_v4_flash_dspark/decode_fwd.py
```diff
@@ -0,0 +1,3348 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# -----------------------------------------------------------------------------------------------------------
+# ci: devices=2  # CI: EP2/TP2 full decode forward
+"""DeepSeek-V4 D-Spark decode forward."""
```
zhangqi-chen, line 11: format the file using /fmt-coding-style

zhangqi-chen (changes requested):

zhangqi-chen (approved):

PR #1069, opened 2026-08-28
Add: support arbitrary DSpark prefill sequence lengths via CP padding

MaxwellF1:
- Support logical DSpark prefill sequence lengths from 1 through 8192
  by executing over a caller-padded physical token extent aligned to
  `TP_SIZE`.
- Define the serving/model-runner contract with zero-padded hidden and
  token-ID inputs, non-aliasing synthetic positions, and `-1` padded
  cache/state mappings; `l3_prefill_fwd` consumes this physical extent
  directly.
- Execute incomplete final 128-row MoE waves with inactive gate outputs
  cleared and staging/output writes bounded to physical rows.
- Keep live logit-row indices within the logical token range and select
  the logical final token for LM-head output.


MaxwellF1:
Serving integration contract

This PR makes the DSpark prefill path work for any logical length N in [1, 8192] through a caller-padded physical ABI. The device entry does not allocate padding from raw N-row tensors. As in vLLM Ascend, the serving/model-runner boundary prepares the physical tensors before dispatch.

- Compute P = align_up(max_group_N, TP_SIZE) and L = P / TP_SIZE. All ranks in one dispatch use the same P/L.
- Assign one request to each contiguous TP group. Replicate the full x_hc sequence within the group, then split rank-local input_ids and position_ids_local by rows [tp_rank * L, (tp_rank + 1) * L).
- Zero padded x_hc and input_ids rows.
- Use monotonic, non-aliasing synthetic future positions for padding. Do not reuse the current MTP repeat-last-position helper: this kernel has no logical-N attention sideband.
- Set the padded suffix of ori, HCA, CSA, indexer, and compressor-state slot mappings to -1.
- Initialize logit_row_indices to -1 and set only the TP-group leader live row to N - 1.
- Read logits and sampled_ids from the group leader. Crop x_out to [:N] before passing hidden states to another serving stage.

The existing pypto-serving MTP padding helpers cannot be reused as-is because they repeat hidden rows, token IDs, and the final position. A DSpark-specific, group-aware staging path is still required in pypto-serving; the attention, HC, MoE, LM-head, and sampling computation remains in pypto-lib.

PR #1076, opened 2026-08-29
Perf: restore DeepSeek V4 MTP blocking waits

MaxwellF1:
- Restore blocking waits for MoE payload/combine and LM-head
  hidden/logits exchange
- Reintroduce producer TaskIds, anchor reads, and explicit dependencies
- Publish MoE completion after the complete local scatter grid

EP8/TP4 decode-logits fastest-rank mean 36.166 -> 35.830 ms,
averaged over two controlled A/B pairs (a2a3, 43 layers, 8 tokens
per rank, 5 warmup + 100 rounds).


MaxwellF1:
## Performance validation

The comparison used the fixed EPLB decode-logits case:

- a2a3, EP8 / TP4
- 43 forward layers, 8 tokens per rank, start position 8192
- devices 0,2,4,6,8,10,12,14
- seed 1807
- 5 warmup + 100 measured rounds
- the rank with the smallest per-rank median was selected, and the table
  reports that rank's mean

| Pair | Blocking wait + manual deps | Deferred wait | Difference |
|---|---:|---:|---:|
| A/B 1 | 35.894 ms | 36.169 ms | +0.275 ms (+0.77%) |
| A/B 2 | 35.765 ms | 36.162 ms | +0.397 ms (+1.11%) |
| Mean | 35.830 ms | 36.166 ms | +0.336 ms (+0.94%) |

Both variants completed all measured rounds without a timeout or deadlock.
This performance entry has no golden function, so these runs validate
execution and latency rather than numerical correctness.

The controlled result is consistent with the independent weekly EPLB
observation of 35.42935 ms -> 36.00040 ms (+1.61%).

## Complete MoE-stage timing

A follow-up device-side timing run used the EPLB decode-logits workload from
`perf/dsv4-eplb-decode-logits-and-mtp-core` at
`25d4f9cf90f28ec8d4108702894d5ce1c9f9c080`.

The earlier wait-to-consumer windows were too narrow for cross-stage
comparison because some producer tasks can be dispatched early. Their
timestamps therefore included scheduler queue residency and could not be
added across layers.

The revised measurement covers a complete representative MoE stage, from
the completion of the attention-side `hc_post` to the completion of the
following MoE `hc_post`. This includes gate and shared-expert work, dispatch,
routed experts, combine, and the associated scheduling gaps.

The controlled A/B used EP8/TP4, 43 layers, 8 tokens per rank, start position
8192, and 5 warmup + 100 measured rounds. The table aggregates the samples
from all eight ranks.

| Complete MoE stage | Blocking wait + manual deps | Deferred wait | Mean difference |
|---|---:|---:|---:|
| CSA-MoE | 564.268 us | 600.030 us | +35.763 us (+6.34%) |
| HCA-MoE | 561.182 us | 572.962 us | +11.780 us (+2.10%) |
| SWA-MoE | 555.892 us | 565.927 us | +10.035 us (+1.81%) |

The corresponding median increases were 2.78% for CSA-MoE, 1.89% for
HCA-MoE, and 1.79% for SWA-MoE. The larger CSA mean indicates additional
tail sensitivity rather than a large intrinsic difference between CSA and
HCA MoE compute: with blocking waits, their complete-stage means differ by
only about 0.55%.

These representative stage intervals may overlap with other scheduled work
and must not be multiplied by the number of layers. They localize the
regression to the MoE communication path but are not an exact decomposition
of the endpoint latency. The uninstrumented endpoint A/B above remains the
primary performance result: restoring blocking waits reduces mean latency
by 0.336 ms (0.94%).

MaxwellF1:
## Post-merge timing correction

This follow-up supersedes the complete MoE-stage timing table in the
[earlier performance comment](https://github.com/hw-native-sys/pypto-lib/pull/1076#issuecomment-5459853786).
The endpoint A/B result in that comment remains unchanged.

The original single instrumented run reported a 6.34% CSA-MoE increase, but
that mean was dominated by a non-reproducible tail: 14 of 800 deferred CSA
samples exceeded 1 ms, with a 2.505 ms maximum. Three additional paired runs
reused the same source commit, compiled artifacts, devices, input, and
5-warmup + 100-measured-round configuration. None of the 2,400 new deferred
CSA samples exceeded 1 ms.

The CSA-MoE mean differences in the three new pairs were:

| Repeat | Blocking wait + manual deps | Deferred wait | Difference |
|---|---:|---:|---:|
| 1 | 570.341 us | 576.329 us | +5.988 us (+1.05%) |
| 2 | 567.099 us | 575.453 us | +8.354 us (+1.47%) |
| 3 | 567.054 us | 575.744 us | +8.690 us (+1.53%) |

Pooling all three repeats across all eight ranks gives:

| Complete MoE stage | Blocking wait + manual deps | Deferred wait | Mean difference |
|---|---:|---:|---:|
| CSA-MoE | 568.165 us | 575.842 us | +7.677 us (+1.35%) |
| HCA-MoE | 564.264 us | 569.136 us | +4.872 us (+0.86%) |
| SWA-MoE | 557.432 us | 563.162 us | +5.730 us (+1.03%) |

All six new 100-round runs completed without a timeout or deadlock. The
repeatable result is therefore an approximately 1% complete-stage regression
across CSA, HCA, and SWA MoE, not a stable CSA-specific 6% regression or tail
effect. This is consistent in magnitude with the uninstrumented endpoint A/B
result of +0.336 ms (+0.94%), while the representative stage intervals should
still not be multiplied across layers as an exact endpoint decomposition.

PR #1095, opened 2026-08-31
Add: support packed multi-request DSpark prefill

MaxwellF1:
- Add packed query boundaries and rank-local request ownership to the
  DSpark DSA-CP prefill ABI.
- Select request-scoped cache, compressor-state, indexer, and sparse
  attention rows across SWA, HCA, and CSA.
- Preserve TP-aligned padding and empty-rank behavior while serializing
  HCA streaming work by request.
- Keep indexer state publication ordered across physical tiles without
  treating fence values as row-validity predicates.
- Thread packed-request metadata through the single-layer and 43-layer
  forward paths, with fixtures that cross TP-rank request boundaries.
- Take an ordered packed request list per TP group, described by
  `query_start_loc = [0, q0, q0 + q1, ..., N]`, where `N` is the logical
  current-chunk token count and must not exceed 8192.
- Expect the packed group padded once to `P = align_up(N, TP_SIZE)`, for
  `L = P / TP_SIZE` rank-local rows; padding rows carry zero
  hidden/input, non-aliasing synthetic positions, and `-1` cache/state
  slot mappings.
- Give the seven cache, indexer, and compressor-state block tables a
  request dimension of `len(query_start_loc) - 1`, with request order
  and metadata identical across ranks of a TP group.
- Emit one live last-token logit row per request on the TP-group leader,
  in request order, and derive rank-local request ownership inside the
  kernel rather than from a serving-supplied local request ID.

Long-prefix chunk scheduling and persistent paged-cache allocation stay
on the serving side; this change does not add mixed prefill/decode
execution to the DSpark kernel.


MaxwellF1:
Validation (PTOAS 0.57, a2a3):

- TP2 SWA, HCA, and CSA leaf golden passed for a ragged B64 batch
  with 161 logical and 162 TP-aligned physical tokens.
- Strict CSA state comparisons passed for B2 and B64 with zero
  mismatches.
- TP2/EP2 43-layer B64 full-forward smoke passed all registered output
  comparators.
- B1 x 8192 indexer-compressor golden passed, covering sixteen 512-row
  state tiles and ring reuse.
- The standalone metadata entry compiled successfully; Ruff, F-rule
  checks with noqa disabled, py_compile, and diff-check passed.

B64 validates packed-request ownership and boundary handling. B1 x 8192
independently validates the long-chunk multi-tile path; this does not claim
a B64 x 8192 workload.

PR #1104, opened 2026-09-01
Fix: preserve inactive prefill indexer score rows

indigo1973:
The golden score buffer is initialized with FP32_NEG_INF for all T
  rows, but only the first num_tokens rows are active. Assigning scores
  to every row overwrote the sentinel values of inactive rows when
  num_tokens < T, causing prefill_score_compare to report false golden
  mismatches.

  Limit both the source and destination to the active token prefix so
  inactive score rows retain FP32_NEG_INF.

  Before the fix, start_pos=0 and num_tokens=8 failed score validation
  with 240 inactive sentinel mismatches:
  - task_20260901_170618_126348616272

  Tested on Ascend A5 in the wzh environment with:
  - PyPTO: 4054785f520cb093d877038f3063a259d3afc3e3
  - runtime/simpler: dbdd041e957420ea15b03e878400dd4de5e9c34c
  - PTOAS: 0.60
  - newer PTO-ISA: ad509ef4aaa078e0db3f89f9e760b27578099ee7

  Test results:
  - start_pos=0, num_tokens=8: PASS task_20260901_172411_133338029446
  - start_pos=0, num_tokens=128: PASS task_20260901_172455_134010821348


lwDavid (approved):
Reviewed.

PR #1108, opened 2026-09-01
CI: fix the CANN pypto shadow and the reset artifact upload

luohuan19:
Two failures in the self-hosted runner scaffolding.

CANN's set_env.sh prepends $CANN_ROOT/python/site-packages to
PYTHONPATH, and CANN 9.2 ships its own pypto package there. PYTHONPATH
outranks a venv's site-packages, so on those hosts that copy shadows the
pypto the job just built: it tries to build its native impl online and
dies on a pybind11 the venv has no reason to carry.

Separately, roughly one nightly artifact upload in eight dies ~3s in
with "Unable to make request: ECONNRESET", on either twirp call
(CreateArtifact or FinalizeArtifact) and on any of the hosts; the blob
PUT between them succeeds, so this is not a size or endpoint-access
limit. The artifact client retries HTTP 5xx five times but re-throws a
raw socket error before that loop (NetworkError.isNetworkErrorCode), so
the failing call is never tried twice.

- activate.sh now rebuilds PYTHONPATH with the venv's site-packages
  first on every source. Rebuilt rather than appended once: each
  task-submit child sources activate.sh again and set_env.sh
  re-prepends its own entry every time, so a skip-if-present guard
  would let CANN back in front from the second source on.
- The post-install check asserts pypto and simpler resolve inside the
  job venv. Importing alone passed with the wrong revision, which is
  how a CANN copy reached the traceback in the first place.
- Add an upload-results composite action: upload, and on an empty
  artifact-id wait 15s and upload once more with overwrite, since a
  Finalize whose response was lost still leaves the artifact created.
  The four daily_ci upload sites use it, keeping their gating on the
  sweep having started, and a second failure still fails the job.
- installation.md gains the CANN symptom for developers who source
  set_env.sh into their own environment.


lwDavid (approved):
Reviewed.

PR #1110, opened 2026-09-02
Perf: derive the hc_pre mix gate inside mix_x

lwDavid:
split_pre_post computed two unrelated things: the pre sigmoid gate, read
only by mix_x, and post, read only by hc_post at the end of the
sublayer. Staging the gate through pre_val_store put a one-block task on
the critical path between hc_pre_rms and mix_x, and all four mix_x
blocks then read the same [T_TILE, HC_PAD] tile back out of GM.

mix_x now derives the gate from mixes_partials and inv_rms directly. The
gate is eight rows of HC_PAD, so the four blocks recomputing it cost a
handful of vector ops, and mix_x depends on hc_pre_rms and hc_pre_linear
instead of on a task that waits for both. What remains of the split is
the post half, which runs off the critical path.

Measured on a5 with the flash variant, 100 rounds, both arms interleaved
on one card so they share machine state:

  decode_attention_swa  326.4 -> 315.6 us  (-3.3%)
  decode_attention_csa  462.2 -> 451.4 us  (-2.4%)
  decode_attention_hca  368.2 -> 362.4 us  (-1.6%)

The swa and csa sample sets do not overlap; hca's do, so its figure is
the least certain of the three. A level-2 chip swimlane confirms the
mechanism: decode_attention_swa's observed critical path drops from 14
tasks to 13 with split_pre_post no longer on it, and
decode_attention_csa's makespan falls from 516 to 487 us with scheduling
stall down from 283 to 250 us.

Goldens pass for every program that shares hc_pre: hc_pre, moe, and
prefill_attention_swa, _csa and _hca.


lwDavid:
### Also measured on the way here, and rejected

Recording these so nobody re-derives them. All A/B'd the same way as the
change itself — both arms interleaved in one job on one card, 100 rounds,
comparing `mean=`.

| candidate | predicted | measured | verdict |
|---|---|---|---|
| `QPROJ_BLOCK_TILE` 8 → 16 blocks | swa −3 to −8 us | **+0.8 us (+0.26%)** | wash |
| hca `GATHER_SEG_TILE` 4 → 8 | hca −6 to −9 us | **+26.2 us (+7.15%)** | regression |
| fuse `score_mat` + `score_reduce` | csa −10 to −20 us | breaks `x_out` | unsafe |

**qproj width** does change the launch as intended (flash 8 → 16 blocks,
`pro` untouched at 16 if written as `((H * HEAD_DIM) // QPROJ_MM_N_TILE) // 16`),
so block count simply is not what binds there. 3 base / 4 variant samples.

**hca gather**: four variant samples all above four base samples. Doubling
to 64 AIV blocks of half the work costs far more than the imbalance it
removes.

**score_mat + score_reduce** is the one worth flagging. The two launches
have byte-identical iteration spaces and block *L* of the first writes
exactly what block *L* of the second reads back through a 4 MB INT32
`score_acc_gm`, so folding the reduce into the matmul task looks free.
It is not: csa `x_out` fails at ~25% of elements over the 0.004 threshold,
deterministically, in all three forms I tried — plain splice; plus an
explicit `weights_tid` dependency on `score_mat` (needed because
`weights_proj` publishes behind the caller's `late_dep`, which is
`deps=[]` standalone but `deps=[rms_tid]` inside `decode_attention_csa`);
and plus de-pipelining the fused loop to `pl.range`.

Two things follow that are worth having on the record:

1. **`decode_indexer`'s own goldens cannot gate this class of change.**
   The standalone case passes everything including `score` and
   `topk_idxs`, because `decode_topk_pair_compare` tolerates tie-flips —
   and a tie-flip changes which KV blocks attention reads, which moves
   `x_out` a long way downstream.
2. It is the same signature as the two intra-task cross-lane handoffs
   already known to break here (folding `quant` into `proj_b_mm`, 12.3%
   of `x_out`; folding the merge into `qk_pv`'s pipeline, 18.5%). I would
   treat an AIC → AIV handoff inside one task as unsafe until someone
   establishes the fence discipline for it — `qr_head` doing vector →
   matmul → vector is not sufficient evidence that the general pattern is
   safe.

### Where csa's critical path went

This change does not shorten csa's path, it re-routes it. The indexer
branch that used to bind (`qr_proj_matmul` → `qr_rms_norm_quant` →
`idx_qr_proj_matmul` → `qr_head`) goes slack and the compressor branch
surfaces: `kv_score_proj_0` at **44.3 us** of compute is now the largest
single item on the path, ahead of `qk_pv`'s 39.1, and it was not on the
path at all before. The four tasks behind it (`scatter_softmax_pool_0`,
`rmsnorm_rope`, `kv_hadamard`, `kv_and_cache_write`) carry 17.6 us of
compute against 33 us of stall — the same staging shape this PR
exploits — but collapsing them needs the AIC → AIV handoff above.

PR #1112, opened 2026-09-02
Perf: drop redundant dependency edges from the SWA decode graph

Little-oil:
An edge A -> C is redundant when the graph already holds A -> B -> C, so
it constrains nothing and only costs the orchestrator a fanout entry and
a released-producer decrement. Three scratch tensors carried such
auto-tracked edges; opt each out of OverlapMap tracking:

- post_t: split_pre_post -> hc_post, covered because hc_pre also feeds
  x_mixed, the head of the chain that ends in attn_out
- qr_scale_pad_store: qr_rms_norm_quant -> qproj_dequant_rms_nope_rope,
  covered by qr_i8 -> qproj_matmul -> dequant
- own_scale: tp_o_a_quant -> tp_o_b_dequant, covered by own_a_i8 ->
  tp_o_b -> dequant; eight edges over four groups and two owners

Six redundant edges on the cross-rank windows are kept, because each is
covered only by a path through a wait task and every wait in this layer
polls peer slots alone:

- the all-gather runs a two-phase epoch on gather_signal: push puts rows
  into every rank including itself and posts a credit, readback copies
  the window out and posts a second, retire resets with AtomicAdd -2
- o_group_a2a and o_proj_reduce_scatter count one credit per publish
  worker, then a completion task notifies once more to close the epoch

A publisher also puts its own rows into its own window, and no peer
credit speaks to that local put, so the retained edge is the only thing
ordering it. They are DistributedTensor parameters besides, with no
manual_dep surface, and two carry duplicate tensormap annotations that
would outlive dropping the explicit deps= entry.

sparse_bias is deliberately left tracked: swa_valid_bias no longer
writes kv_cache, so nothing else orders it ahead of qk_pv.

Add --enable-dep-gen to match the sibling entries; the audit needs it.

Generated kernel code is unchanged: stripped of comments, the only
difference across all 41 kernels is SSA numbering, so buffer packing and
tiling are untouched and the whole delta is orchestration wiring.

Wall time is unchanged, so this is not a speedup. It is kept because it
removes constraints the follow-up allow_early_resolve work needs, and
because the graph is smaller. Measured over three alternating pairs of
100 timed rounds on a2a3, TP=4, cards 3,5,7,9, 16 requests at start-pos
8192, swimlane off: max-rank effective_us median 2490.2 -> 2487.1 us
(-0.13%), mean 2502.8 -> 2704.3 us (+8.0%). The mean difference is 0.14x
the same-version noise floor, which reached 763.7 us on the baseline arm
and 1461.7 us on the change arm across repeat runs of identical code,
and it reverses sign under the median, so it is not a regression either.
One open observation, not established at n=3: the change arm showed
round-level spikes above 1.5x its own median in 2/19/1 rounds per run
against 0/0/0 for the baseline arm, consistent with a greedy scheduler
having more orders to choose from. Graph size is the reproducible
effect: l3_decode_swa rank0 223 -> 213 task edges and 16 -> 6 redundant,
63 tasks unchanged.


Little-oil:
CI triage: both `sim` failures are pre-existing on `main` and unrelated to this PR. The `a2a3` device job, the only one that executes these kernels on hardware, passes.

**`sim (a5sim)`** fails in `AllocateMemoryAddr`:

```
ValueError: Right buffer usage (98304 bytes) exceeds platform limit (65536 bytes)
  at models/deepseek_v4_flash_dspark/qkv_proj_rope.py:375:13
```

Reproduced locally on unmodified `upstream/main`, byte counts and site identical:

```bash
git checkout upstream/main -- models/deepseek_v4_flash_dspark/qkv_proj_rope.py
python models/deepseek_v4_flash_dspark/qkv_proj_rope.py -p a5sim --compile-only
```

It fails 11 files across the whole dspark family, including `dspark_prefill.py` and `dspark_context_kv.py`, which this PR does not touch. The site is in the `kv_proj` region reworked by #1103.

**`sim (a2a3sim)`** hangs in `decode_hca` after `decode_csa` passes:

```
TIMEOUT_EXIT sub_class=S1:running-stalled completed=73/111 running=1 ready=1
waiting=36 orch_done=1 stuck_task_id=12884901908
```

Reproduced locally on unmodified `upstream/main`: `decode_hca.py -p a2a3sim` compiles, computes golden, enters `runtime`, and never returns (killed at 900 s).

**Same signature on merged PRs.** #1102 and #1103 both show `sim (a5sim) fail` and `sim (a2a3sim) fail 30m22s`, the same 30-minute timeout, with `a2a3 pass`. This PR reproduces that combination exactly.

Both look worth their own issues; neither is addressed here.

**What this PR was verified against:** a2a3, devices 4 and 6. Accuracy passes for `kv_cache` and `x_out`. The dependency graph goes from 223 to 213 unique task edges and 16 to 6 redundant edges over 63 unchanged tasks, rank0 and rank1 symmetric, measured from `--enable-dep-gen` captures before and after.

PR #1117, opened 2026-09-02
Perf: optimize DeepSeek V4 Pro decode indexer

indigo1973:
Fuse scatter_softmax_pool, rmsnorm_rope, and kv_hadamard into scatter_pool_rmsnorm_rope_hadamard. Keep pooled_kv and normed_kv local to the fused MIX kernel to remove GM round trips and task boundaries, while keeping kv_and_cache_write separate.

Fuse score_mat and score_reduce into one split MIX score kernel. Consume the INT8 matmul accumulator locally, perform ReLU, head reduction, and dequantization in-kernel, remove score_acc_gm, and make topk depend on the fused score task.

Keep the default dependency-resolution policy with no allow_early_resolve overrides. Extend --enable-chip-swimlane choices through level 4. The golden implementation is unchanged.

A5 benchmark on device 1 (PYPTO_BENCH=1, 5 warmups + 50 measured rounds): parent baseline effective_us min/median/mean/max = 124.8/132.0/133.9/150.7 us; optimized = 100.5/112.2/113.3/133.0 us. Mean improves by 20.6 us (15.38%, 1.182x). All output checks PASS.


lwDavid (approved):
Reviewed.

PR #1118, opened 2026-09-03
Perf: parallelize the CSA decode compressor's serial scopes

Little-oil:
- Run scatter_softmax_pool and compress_state_commit as pl.spmd(b_dim)
  rather than one pl.at(level=CORE_GROUP) block, in both
  decode_compressor_ratio4 and decode_indexer_compressor: each block
  takes its request from the block index, and the bodies are unchanged
- Re-nest compress_state_commit's flat loop over tokens as request x
  position so it matches the new block decomposition

Both nests were already parallel over requests: a request writes only
its own pooled_kv rows and its own state-ring rows -- whole rows of
2 KB and 8 KB, so no 64-byte line is shared between blocks -- reads its
ring through its own compress_state_block_table row, and carries
nothing across requests. compress_state_commit additionally relies on
S = 8 <= STATE_LEN = 8, which puts a request's tokens on distinct
positions mod STATE_LEN and distinct state pages per request; the
serial loop relied on last-writer-wins instead.

The four scopes held 3.4 ms of single-core work inside a 4.5 ms layer,
and the cost was not only their own runtime. indexer_score_leaf_wave is
a 24-block MIX task, so the cluster whose AIVs a serial scope occupied
could not start until the scope released, with its AIC half waiting
inside the kernel; its other 23 clusters ran 273-348 us each.
indexer_topk_group_wave lost the same way to compress_state_commit.

Measured on a2a3, TP=4, cards 0,2,4,6, 16 requests at start-pos 8192,
100 timed rounds after 5 warmup, golden recomputed each run, five
before/after pairs run alternately. Fastest-rank mean drops
4241.4 -> 2375.3 us, -44.0 % at the median of 5 (per-run 4201-4353
before, 2306-2715 after); the max-rank effective_us headline drops
6881.9 -> 4936.3 us, -28.3 %, the gap being cross-rank all-gather skew
this change does not touch. A level-4 chip swimlane attributes it:
scatter_softmax_pool 2121.0 -> 74.3 us across 1 -> 64 blocks, its
indexer twin 596.8 -> 30.7, compress_state_commit 332.4 -> 34.2 and
304.9 -> 36.9, indexer_score_leaf_wave 1890.9 -> 354.0, layer span
4450.8 -> 2444.4, with untouched qk_pv flat at 478.7 -> 480.9. Core
busy time is unchanged: scatter_softmax_pool costs 2121 core-us on one
core before and 2248 across 64 blocks after.


Little-oil:
修改前：
image
修改后：
image

PR #1125, opened 2026-09-03
Fix: take pypto's DFX options by its public name

lwDavid:
## The problem

`golden/runner.py` imported pypto's DFX options bundle by its private
name, `pypto.runtime.runner._DfxOpts`. pypto has since exported the same
dataclass publicly as `pypto.runtime.DfxOptions` and dropped the private
alias, so against current pypto `main` any run that sets one of the five
`_DFX_FLAG_KEYS` fails before it starts:

```
ValueError: This pypto runtime does not support execute_compiled
DFX flags: ['enable_chip_swimlane']
```

That is the whole DFX surface, not a single flag:

| flag | what it takes out |
|---|---|
| `enable_chip_swimlane` | every swimlane capture, at any level |
| `enable_dep_gen` | `hc_pre`, which sets it unconditionally |
| `enable_pmu` | PMU counters |
| `enable_dump_args` | argument dumps |
| `enable_scope_stats` | scope statistics |

`hc_pre` has no way out: it sets `enable_dep_gen` for the `separate`
impl, a5 rejects `--impl syncall` outright, and `--no-dep-gen` is a
documented no-op retained only for CLI compatibility.

## The fix

Reaching for a leading-underscore name in another package is ours to
fix, not upstream's, so this takes the public name and keeps the private
one as a fallback for pins that predate the export. Both bind the same
five keyword arguments, so either satisfies the harness. When a runtime
carries neither, the caller still gets the same `ValueError` it got
before rather than an `AttributeError`.

Only `golden/runner.py` needed it. The two other private-pypto imports
in the tree — `_to_torch_dtype` and `_to_runtime_shape`, in
`golden/runner.py` and `models/deepseek_v4_pro/synthetic_token_loop.py`
— still resolve on current pypto and have no public equivalent to move
to, so they are left alone.

## Validation

On 8p (Ascend 950PR) against pypto `b8165168`, simpler `77fa0171`,
pto-isa `96ba706c`, ptoas v0.57 — the toolchain on which both paths were
dead before this change:

```
hc_pre --variant flash -p a5
    x_mixed / post / comb PASS
    dfx_outputs/deps.json written

decode_attention_swa --start-pos 8192 --enable-chip-swimlane 1
    TOTAL 233 scopes / 3242.96 core-us
    AICore Observed Span 324.81 us
    swimlane JSON written
```

`hc_pre` itself is unchanged at ~50 us — 49.6 / 50.1 / 50.0 on this
toolchain against 51.3 / 51.1 / 49.9 on one that still carries
`_DfxOpts`, three repeats interleaved arm-by-arm on a single card. The
published figure for it is 49.7.

`tests/golden/test_runner.py` now covers all three cases: the public
name, the fallback to the private one, and a runtime carrying neither.
`pytest tests/golden` — 303 passed. `ruff check` clean.

Note that the repository's `ruff format` is not clean on these two files
at `main`, so this branch deliberately does not run it; the diff is only
the logical change.


lwDavid:
Superseded by #1128, which carries the same fix.

#1128 makes `_execute_compiled_kwargs` import `pypto.runtime.DfxOptions` and fall back to `runner._DfxOpts` on an older pypto, which is exactly the change here, and it adapts the rest of pypto's consolidated compile/execute entry points at the same time. Keeping both open would only put two PRs on the same few lines of `golden/runner.py`.

Closing in favour of #1128. Worth landing soon either way: until one of them merges, `--enable-chip-swimlane` and the other `runtime_cfg` DFX flags are unusable on current pypto, so the DSv4 model entry points cannot be profiled at all.

PR #1126, opened 2026-09-04
Perf: run the DSpark CSA decode SPMD launches on persistent workers

Little-oil:
- Convert fifteen SPMD launches across decode_compressor_ratio4,
  decode_csa, decode_indexer, decode_indexer_compressor and
  qkv_proj_rope from one block per work item to a fixed set of
  persistent workers striding over the same block list.
- Hoist the loop-invariant operands each block was rebuilding to once
  per worker: the 32 KiB hadamard matrix in both hadamard matmuls, the
  rope swap index in qr_rope, and the arange-derived interleave lane,
  swap index and sign fold in q_rope_prepare, whose core time halves
  from 500.0 to 248.6 core-us.
- Swap kv_hadamard's column-tile loop outside its row-block loop so a
  worker reuses one hadamard slice across the row blocks it owns.
- Cap q_rope_prepare at pl.min(Q_ROPE_WORKERS, token_tiles) so its
  16-tile call site does not dispatch more blocks than it did before.
- Leave work items, tiling, memory traffic and results unchanged: every
  tile constant, valid_shape and divisibility requirement stays where
  it was.

The layer handed the AICPU scheduler 3006 blocks per step at about
1.02 us of dispatch plus completion each, with its three threads at
81.6% occupancy against AIC at 71.4% and AIV at 10.7% -- dispatch
budget, not cores, is the constraint. The layer now issues 1750 blocks
(-42%), and the converted launches drop from 1811.5 to 530.4 us of
dispatch. Fastest-rank median wall time improved from 2348.6 to
2233.3 us (-4.9%) over five alternating rounds against one frozen
golden (a2a3 devices 0/2/4/6, TP4, 16 requests at 8K, 100 rounds /
5 warmup). The gain saturates once the first three launches convert;
the later ones measure within noise and are kept for the dispatch
budget and per-block setup they return, not as speedups. Worker counts
are defensible defaults rather than swept values -- 24 for cube tasks,
48 for vector -- and several launches sit at 1.33 work items per
worker, which a follow-up can size down. decode_csa validation passes
all seven checks after each commit, and decode_swa is validated on
device for the shared qkv_proj_rope change. Supersedes #1124.


Little-oil:
改前
image

改后
image

Little-oil:
下一步调顺序

PR #1129, opened 2026-09-04
Perf: widen the DSpark compressor's cube M-block

wangqin1723-max:
- Give kv_score_proj 64-row M-blocks instead of the 16-row fractal
  minimum, cutting its SPMD block count from 256 to 64.
- The task was dispatch-issue and weight-load bound rather than M bound:
  256 blocks of 10.9 us tiled a 114 us compute floor inside a 305 us
  wall, and the weight tiles reload once per M-block. Widening M to 32
  raised per-block exec only 10% while cutting the task's core time 45%
  and its wall to 192 us, so the step to 64 follows the same slope.

Fastest-rank median 1165.8 -> 1064.2 us, -8.7% (a2a3, TP=4, batch 16
uniform 8K, 100 rounds, interleaved against main with a frozen golden
replayed).


wangqin1723-max:
Folded into #1131, which carries this M-block widening plus two more block-count cuts and a combined measurement against main.

PR #1131, opened 2026-09-04
Perf: cut the DSpark HCA decode SPMD block counts

wangqin1723-max:
- Widen the compressor's cube M-block from 16 to 64 rows, cutting
  kv_score_proj from 256 SPMD blocks to 64. Its 10.9 us blocks tiled a
  114 us compute floor inside a 305 us wall. Widening M also cuts the
  task's MTE2 weight traffic, since the weight tiles reload once per
  M-block: 16 -> 32 raised per-block exec only 10% while cutting core
  time 45% and the wall 305 -> 192 us.
- Run q_rope_prepare on 24 workers instead of 48. The task holds only
  ~273 core-us, so its 48 blocks bought a single 5.7 us wave for ~26 us
  of serial block dispatch.
- Give the HCA cache writeback 16-token blocks instead of 8, halving it
  from 64 blocks to 32. The 64 blocks needed two AIV waves for the same
  7.9 us of work that 32 fit into one, at twice the dispatch cost.

Fastest-rank median 1127.6 -> 1023.4 us, -9.2%, with all four completed
paired rounds negative (a2a3, TP=4, batch 16 uniform 8K, 100 rounds,
interleaved against main with a frozen golden replayed).


wangqin1723-max:
Closing: the MM_B_TILE 16 -> 64 widening that carried this PR's measurement is rejected by the current pypto's AccCompactValid verification (decode_compressor_ratio128.py:163/164), which the older compiler I measured on did not enforce. A sweep on the current compiler shows only MM_B_TILE = 16 compiles; 32, 48 and 64 all fail. The two remaining constants (Q_ROPE_WORKERS, HCA_WB_TOKEN_TILE) show no demonstrable benefit against current main once the widening is removed: -3.9 us mean over five interleaved rounds with inconsistent direction.

Issue #1133, opened 2026-09-04
[Bug][Tracking] DeepSeek V4 persistent prefill/decode correctness gaps


sjduan:
## Background

This tracks the remaining correctness and validation work for DeepSeek V4
D-Spark persistent prefill/decode. The initial deployment target is
DP4/TP4/EP16, with a fixed decode capacity of 64 requests per TP group and
S=8. Serving decides how many tokens to accept; persistent KV/state writes
must remain safe when only the first token is accepted.

This is a model/serving integration tracker, not a claim of a newly
reproduced compiler defect. Confirmed source/host findings, numerical
differences, and uncompleted device gates are distinguished below.
Passing one standalone forward does not close the multi-step gate.

### Environment and source baseline

- Current lib audit baseline: 774b46c38f62d930050f7fc468e42acb24118c51.
- Earlier evidence used lib 4b363644763fdfe2926cdbd987343ec473657d05
  with pending D-Spark transaction/occupancy/RoPE fixes.
- Serving adapter audit: 79f6fd4b93aeaf7f51185e8ab2a125c1e1a64ac2 plus
  matching, not-yet-upstreamed adapter fixes. Serving findings below do not
  imply that their implementation lives in pypto-lib.
- Historical device environment: a2a3; PyPTO
  9531c745d173f919928650cec74555f5b006258a; simpler
  15f5cbd922494c62444e75fa1585e39f86a64a78; PTO ISA
  3e61cd46a55a047234d1ce41856584624f098550; PTOAS 0.57.
  CANN version was not recorded. This is not a new pin-conformance claim.
- Host metadata comparisons used CPU PyTorch on Linux (aarch64).
  No new device run is claimed for the rebased combination.

### Description / remaining work

- [x] **A — P0: Prefill main ratio4 state aliasing.**
  In models/deepseek_v4_flash_dspark/prefill_compressor_ratio4.py,
  PREFILL_STATE_TILE is 512 and prefill_c4_state_scatter_pre writes the
  tile before prefill_c4_softmax_pool reads persistent state.
  The serving request ring has only eight two-row pages (16 rows).
  Positions 0 and 16 therefore alias within one tile. An expanded logical
  page table does not provide additional physical rows. Ordering adjacent
  tiles cannot fix aliasing inside the same tile.
  Protect old history, pool current rows from projection scratch, and
  publish a correct state tail; alternatively define a separate prefill
[...]

PR #1134, opened 2026-09-04
Fix: preserve DSpark decode transaction state capacity

sjduan:
- Reserve compressor state for the historical window plus the full
  decode transaction, with stable per-request physical allocations.
- Keep fixture page tables independent of the active batch size and
  validate packed layer pools against their per-layer geometry.


sjduan:

Serving coordination: the CSA main/inner readers now use `STATE_LEN + S` physical rows (16 at S=8) and eight state-table pages per request. Update allocation, modulo write slots and the scheduler's full-S write budget together; the old four-page serving ABI must not be mixed with this reader. HCA defaults to 1088 state pages per layer; this is not the compressed-KV pool. Small non-multiple-of-four HCA active batches also need the RoPE tail fix in #1136.

PR #1137, opened 2026-09-04
Fix: isolate DSpark parallel cache-line writers

sjduan:
- Stage index KV scales before a single cache writer publishes them.
- Write empty CSA softmax statistics as full head tiles and assign
  each MoE input-ID block one complete INT64 cache line.
- Document and check the alignment of MoE expert-row metadata.


sjduan:

The scale commit returns the final completion TaskId so readers also wait for published scales. The fully-invalid sparse-window scenario is introduced separately by #1139. Validate both fixes together for inactive owners; inspect writer ownership rather than treating all warnings as proof of a race.

PR #1138, opened 2026-09-04
Fix: declare DSpark prefill scratch buffers as outputs

sjduan:
- Mark seven device-written prefill workspaces as outputs in the
  rank-local and L3 interfaces so writes are reflected in the ABI.
- Register the scratch outputs with finite-completion comparators
  instead of comparing their device results against untouched zero
  references. Preserve the existing head, logits, and sampled-id
  validation.


sjduan:

Only parameter directions change here: seven scratch classes at both prefill levels. There is no prefill compressor algorithm change or additional tensor argument. The serving HC embedding/query-length fixes remain a separate cross-repository change, and tracker #1133 still covers unresolved prefill state pooling.

sjduan:
Updated in dbe463f5. The seven newly declared scratch outputs now use the existing finite-completion comparator, rather than default allclose against untouched zero references. The kernel and head/logits/sampled-id oracles and tolerances are unchanged.

Validation:
- The CPU regression rejects the old version for exactly the seven missing comparator registrations; the updated version covers every L2/L3 Out/InOut parameter. Finite nonzero scratch passes; empty buffers and floating-point NaN/Inf are rejected.
- All pre-commit checks and syntax checks pass.
- A2/A3 default prefill_fwd, TP2/EP2, PTOAS 0.57: compile/runtime/validation PASS, including all seven scratch outputs and x_out/logits/sampled_ids. Command: `python models/deepseek_v4_flash_dspark/prefill_fwd.py -p a2a3 --ep 2 --tp 2`, using two scheduler-allocated devices. Result: `[RUN] PASS (191.62s)`.
- GitHub CI run 33859693216 is complete: a2a3 PASS (5m5s), pre-commit PASS, unit-tests PASS. Both simulator jobs pass, but prefill_fwd remains device-only under its existing no-sim marker.

This validates the existing forward topology/completion and terminal oracles, not an independent numerical golden for all 43 layers.

PR #1146, opened 2026-09-07
Perf: serialise the DSpark CSA decode cube launches

Little-oil:
- Chain the six large cube launches head to tail so each gets the whole
  cube array: idx_qr_proj_matmul -> kv_score_proj (indexer) ->
  kv_score_proj (ratio4) -> qr_hadamard_matmul -> kv_hadamard ->
  qproj_matmul
- Fence rmsnorm_rope_cache_write, compress_state_commit, kv_proj_rope
  and csa_cache_writeback behind indexer_score_leaf_wave, a 24-block MIX
  launch that loses a whole wave to a single task squatting on a cluster
- Split indexer, indexer_compressor, compressor_ratio4 and q_proj_rope
  into halves so the ordering is expressible, the producers having sat
  after their new consumers in program order; every wrapper keeps its
  signature, leaving decode_csa_tp1 and the standalone tests untouched
- Move indexer_score_leaf_wave and score_arena out of the top-k arena
  scope so the leaf TaskId survives the scope close
- Run qproj_matmul (64 blocks) and weights_proj (32 blocks) on 24
  persistent workers, one per AIC, instead of fragmenting into small
  waves that reload their weights once per wave
- Drop two dummy barrier tasks whose fences a data dependency already
  implies

Decode step 2241 -> 2140 us (paired-difference median -100.7 us,
-4.50%, negative in all five rounds) on a2a3, TP=4, 16 requests at
start-pos 8192, 100 rounds / 5 warmup, fastest rank.


Little-oil:
改前：
image
改后：
image

PR #1151, opened 2026-09-07
CI: add Pro a5sim daily coverage

yanghaoran29:
- Route the a5sim daily model leg to DeepSeek V4-Pro while keeping
  non-Pro models on a2a3sim
- Scope the Pro preset to a5sim and honor existing no-sim exclusions
- Group A5Sim results with A5 device results and label the device-only
  prefill MTP case


.github/workflows/daily_ci.yml
```diff
@@ -71,6 +70,9 @@ jobs:
           PYTHONPATH: ${{ github.workspace }}/checkout
           PYPTO_LOG_LEVEL: error
           PYPTO_WARNING_LEVEL: none
+          # Keep the Pro simulator leg pinned to the Pro preset even if the
+          # runner environment selects the optional Flash preset.
+          DEEPSEEK_V4_VARIANT: pro
```
zhangqi-chen: This is a step-level env, so it applies to the `a2a3sim` leg as well, not just the new `a5sim` one. It is inert there only because `models/deepseek_v4_pro/config.py` is the sole reader and that leg now skips Pro entirely — the pin still says something the leg it lands on cannot act on.

It is also redundant as written: `config.py` resolves the variant with `os.environ.get(VARIANT_ENV, "pro")`, so an unset env already selects Pro. The line only changes behaviour if the runner environment exports `flash`, and in that case it should be scoped to the leg that actually sweeps Pro:

```yaml
DEEPSEEK_V4_VARIANT: ${{ matrix.platform == 'a5sim' && 'pro' || '' }}
```

.github/workflows/daily_ci.yml
```diff
@@ -958,7 +965,7 @@ jobs:
 
           print("## Daily CI Model Test Results")
           print()
-          table("a2a3 and a2a3sim", ["a2a3", "a2a3sim"], "a2a3")
+          table("a2a3 and simulators", ["a2a3", "a2a3sim", "a5sim"], "a2a3")
```
zhangqi-chen: This re-creates the half-filler matrix the two-table split exists to prevent.

After this PR `a5sim` sweeps exactly `models/deepseek_v4_pro` (33 runnable, non-`no-sim` cases) and `a2a3`/`a2a3sim` still exclude it (~48 cases), so the case sets remain disjoint. `table()` prints the union of cases across the present platforms, so the combined table gets ~81 rows in which every Pro row is `-` under `a2a3`/`a2a3sim` and every non-Pro row is `-` under `a5sim` — and the Pro rows then repeat in the `a5` table below. The replacement comment above ("the first table combines simulator correctness ... with a2a3 device coverage") describes the opposite of what the code renders.

`a5sim` and `a5` sweep an identical case set, so they align in one table:

```python
table("a2a3 and a2a3sim", ["a2a3", "a2a3sim"], "a2a3")
table("a5 and a5sim (DeepSeek V4-Pro)", ["a5", "a5sim"], "a5")
```

zhangqi-chen (commented):
Two notes on the workflow changes.

.github/workflows/daily_ci.yml
yanghaoran29: Fixed in `20e54d6` — scoped the Pro preset to the a5sim matrix leg.

.github/workflows/daily_ci.yml
yanghaoran29: Fixed in `20e54d6` — split the summary by model family and label the device-only prefill MTP case as `only a5`.

PR #1153, opened 2026-09-07
Fix DSpark drafter RoPE payload scaling

hashiqiqixian:
### Summary

Replace the DSpark drafter's full 1M-token resident RoPE tables with bounded, token-local RoPE rows for the context and query streams.

This addresses the pypto-lib portion of item L in #1133.

### Problem

The previous L3 drafter ABI carried BF16 cosine and sine tensors shaped `[N_RANKS, MAX_SEQ_LEN, 64]` and marked them resident. At the 1,048,576-token limit, the pair consumes 256 MiB per rank, or 4 GiB across EP16, even though one drafter invocation only reads its bounded context and query rows.

### Changes

- Make `dspark_context_kv` consume the selected context RoPE rows directly.
- Give `dspark_attention` separate selected RoPE streams for owner-local Q rows and full DSA-CP-group KV rows.
- Replace the two full-table drafter inputs with six bounded inputs:
  - context-group cosine/sine: `[CP_CONTEXT_T_DYN, 64]`
  - local-query cosine/sine: `[112, 64]`
  - group-query cosine/sine: `[448, 64]`
- Keep position IDs as metadata for admission, layout validation, and Golden semantics.
- Generate fixture rows with the existing `token_local_rope` helper and consume them with the existing `rope_prepare` implementation. No new operator or communication primitive is introduced.
- Strengthen the component and drafter Goldens so local-Q, group-query KV, and context-group KV use independently observable RoPE rows.

### Memory impact

- Before: 256 MiB/rank resident RoPE payload; 4 GiB across EP16.
- After, at the maximum 8,192-row context: 2,240,512 bytes/rank (about 2.14 MiB), or about 34.2 MiB across EP16.
- The maximum aggregate payload is reduced by about 99.17%, and the six replacement tensors are transient rather than resident.

### Validation

Local checks on `42c8e3e`:

- Python syntax, header lint, public-doc boundary checks, English-only lint, and `git diff --check`: PASS.
- Complete code review and deletion/minimality audit against `origin/main` (`5922ad6`): PASS.

A2/A3 NPU validation on 686 with PyPTO `9f657f37`, Simpler `4e4d3a4`, and PTOAS 0.57:

- `task_20260906_221639_35400219502`: `dspark_context_kv`, Decode T=128 and Prefill T=512, numerical validation PASS.
- `task_20260906_221723_35559914791`: `dspark_attention` at the 16K boundary and 1M tail, KV cache and packed attention outputs PASS.
- `task_20260906_221829_365184614822`: DSpark drafter B8, TP4/EP4, intermediate state, exact CP-group KV replicas, and head hidden PASS.
- `task_20260906_221948_369261828036`: DSpark Prefill B4, TP4/EP4, intermediate state, exact CP-group KV replicas, and head hidden PASS.

### Scope and integration boundary

This PR changes the pypto-lib drafter model ABI only. Serving M2 must replace its static full-table tensors with the six bounded per-dispatch tensors in the same order and regenerate compiled artifacts; old Serving call sites and old compiled artifacts are intentionally ABI-incompatible.

This PR does not close the full #1133 tracker and does not claim a DP4 x TP4 / EP16 Serving end-to-end run. It establishes the bounded lib-side ABI and validates the numerical component and TP4/EP4 model paths.


hashiqiqixian:
Superseded by #1150. The bounded DSpark drafter RoPE patch has been incorporated into the consolidated PR with equivalent code and validation history.

PR #1154, opened 2026-09-07
Fix: serialize DSpark SWA lens metadata writes

hashiqiqixian:
## Problem

`swa_lens` has shape `[3, 16]` and dtype INT32, so each layer occupies
exactly one 64-byte cache line. The previous request-parallel metadata task
let 16 AICore blocks write separate scalars in each shared line. Because a
scalar DDR write commits a complete cache line, those stores could overwrite
one another. The same drafter graph is also used by DSpark Prefill.

## Change

- Generate all 48 SWA-length scalars in the existing single-core metadata
  task, giving every cache line one physical writer.
- Keep the wide `[3, 16, 192]` SWA-index rows request-parallel; each request
  already owns complete, 64-byte-aligned rows.
- Preserve the existing tensor ABI, shapes, visible-window formula, and
  inactive-request semantics.

## Validation

- CI run [34090297795](https://github.com/hw-native-sys/pypto-lib/actions/runs/34090297795):
  pre-commit, unit tests, A2/A3 simulator, A5 simulator, and A2/A3 hardware all
  passed.
- A validation-only device wrapper compared every `swa_lens` value exactly
  for B4 and B16 (`task_20260906_231110_314229819966`). The compiler no longer
  reports `ScalarWriteLineShared` for `swa_lens`.
- The complete TP4/EP4 drafter passed on four A2/A3 devices for both B4 and
  B16, including hidden outputs and exact within-group KV-cache comparison
  (`task_20260906_231345_3347917516`).

This is a model-side writer-ownership fix; it does not by itself establish the
full DP4/TP4/EP16 Serving release gate tracked by #1133.

Refs #1133


hashiqiqixian:
Superseded by #1150. The DSpark swa_lens single-writer patch has been incorporated into the consolidated PR with equivalent code and validation history.

PR #1160, opened 2026-09-07
Fix DSpark prefill MoE ID tail writes

hashiqiqixian:
### Summary

- Make each `prefill_moe_ids_stage_tail` block own eight adjacent `INT64` token IDs, exactly one 64-byte cache line.
- Preserve active token IDs and zero padding for inactive rows.
- Keep the existing full-wave DMA path unchanged.

### Problem

The partial-wave path launched one block per token and used scalar GM writes. Eight concurrent blocks therefore wrote different 8-byte words in the same 64-byte cache line. A whole-line writeback from one core could overwrite neighboring token IDs produced by another core, which could route hash layers to the wrong experts for arbitrary-length Prefill inputs.

The S512 TP4 fixture has exactly 128 local rows and does not execute this branch; Serving inputs with a partial final local wave do.

### Scope

This is a one-file ownership fix. It does not change tensor shapes, public ABI, MoE math, task dependencies, communication, or the full-wave data path.

### Validation

- Full code review and deletion/minimality audit: PASS.
- `git diff --check` and Python syntax: PASS.
- ci-69 compile task `task_20260907_175740_17812712509`: PASS (`exit=0`, 15.91s).
  - Configuration: A2/A3, TP4 / EP4, hash layer 0, 516 group tokens (129 local rows: one full wave plus a one-row tail).
  - The prior `ScalarWriteLineShared` diagnostics for `input_ids_wave_rows` / `prefill_moe_ids_stage_tail` are gone.
  - Toolchain: PyPTO `783259310a4cc0b6bef8bda99186c5508e120d9b`, Simpler `4e4d3a4ad1e54c1db3d50e72decc025a9075bfa0`, PTOAS v0.60, PTO-ISA `a8040450238f162985d8b596fbebeb54bfba2bf5`.
- A2/A3 numerical validation for full-wave and partial-tail cases is queued and will be added here when complete.

Refs #1133


hashiqiqixian:
Superseded by #1150. The DSpark Prefill MoE ID cache-line writer patch has been incorporated into the consolidated PR with equivalent code and validation history.

PR #1167, opened 2026-09-08
Perf: remove redundant DSpark CSA decode computations

Little-oil:
- Pack only ratio-4 compression-boundary rows before the main and indexer
  compressor RMSNorm/RoPE. At TP4, local B16 and S8, each rank processes
  128 gathered boundary rows instead of 512 rows, and avoids the unused
  pooled-zero writes.
- Compute the maximum compressed-cache length once and share it with
  indexer score and top-k workers, replacing their repeated batch scans.
- Reuse the shifted compressed index in attention planning. Simplify the
  sink denominator and cast only the 448 NOPE columns before inverse RoPE;
  the 64 RoPE columns retain FP32 intermediates and their final BF16 cast.
- Preserve packed request indexing on PTOAS 0.60 by avoiding a nested
  divisor that loses parentheses during C++ emission.

At the 16 x 256 test point, matched swimlane-4 captures show both
compressor RMSNorm tasks shrinking from 32 blocks to 8. Main compressor
`rmsnorm_rope_cache_write` execution span falls from 36.36-66.18 us to
16.70-17.14 us across the same four devices (four-rank median:
47.70 -> 16.73 us). Indexer `rmsnorm_rope` reduces core busy time, while
its median task span remains approximately unchanged: 13.69 -> 13.72 us.
Task span means first block start to last block end, including block
scheduling spread; it is not the unprofiled CSA benchmark average.

The `qk_pv` AIC/AIV executable code sections are identical on the measured
toolchain, and its captured duration changes in both directions across
ranks. This is not an established QK/PV compute speedup. The test takes
the single-group top-k path, so it does not time `indexer_topk_group_wave`.
Task overlap and variable communication waits prevent assigning the
whole-CSA mean difference to one task or summing task savings.

Untrimmed CSA means were 1382.011 -> 1321.843 us (4.35% lower) in the
baseline-then-PR pair, and 1407.594 -> 1345.572 us (4.41% lower) in an
additional PR-then-baseline audit. Both pairs support an improvement in
the mean, but contain large timing tails. Selected-rank medians were
1334.100 -> 1319.950 us and 1352.450 -> 1324.600 us respectively;
these are diagnostics, not substitutes for the means. An exact stable
percentage and a per-kernel share of the CSA gain are not established.

Measured baseline `fa5bf7e` -> PR `93f71e3` at 16 local requests per rank,
all with `start-pos=256`, S8, TP4 (64 requests / 512 tokens per group,
128 local tokens), on a2a3 devices 0/2/4/6 with `PYPTO_BENCH=1`,
`--enable-chip-swimlane 0`, 100 measured rounds after 5 warmups per
variant, quoting the lowest of the four per-rank arithmetic means with
all samples retained. Each pair ran under one card allocation and
replayed the same frozen inputs and golden outputs; all seven checks
passed. Both new capture and reversed-order audit launchers exited 0.
PyPTO `f1bb0860`, runtime `15f5cbd9`, PTOAS 0.57, CANN 9.0.0.


Little-oil:
前：
image
后：
image

PR #1187, opened 2026-09-09
Perf: pipeline remaining DSpark QK/PV attention

Little-oil:
- Apply the explicit QK/softmax/PV pipeline from #1183 to DSpark SWA,
  HCA, drafter attention and prefill. Use three bounded transfer slots
  and AIC/AIV events; keep running softmax state in UB.
- Run SWA and HCA raw-window QK/PV over all 64 heads and pipeline
  across queries. HCA raw attention replaces the two 32-head loops
  and split 256-column QK/PV matmuls with full-width matmuls, while
  preserving the raw validity mask and empty-window behavior.
- Make HCA compressed QK/PV query-major: reuse Q across visible
  128-row blocks, skip invalid work, and publish one accumulated
  state per query. Remove per-block neutral stores, shrink m/l
  columns from eight to one, and replace both TP1 and distributed
  compressed-block merge loops with one raw/compressed merge.
- Apply Q reuse, empty-block skipping and online accumulation to
  drafter and prefill attention. Skip unused KV staging and remove
  their downstream block merges. Preserve drafter's 64-row blocks
  and BF16 probability rounding, and streaming HCA's 32-head tasks.
- Add opt-in saved input/golden replay to the standalone attention
  entries.

Measured on a2a3 TP4, 16x256 per rank at S=8, PYPTO_BENCH=1,
swimlane0, 100 timed rounds after 5 warmups, replaying frozen inputs
and goldens, against main 216b497, which already carries #1170 and
#1183. The metric is the minimum per-rank arithmetic mean over every
sample. Toolchain: PyPTO f1bb086, simpler 15f5cbd, PTOAS 0.57, PTO ISA
96ba706, CANN 9.0.0.

HCA decode falls 968.882 -> 753.039 us on devices 0/2/12/14 and
867.430 -> 732.239 us on 0/2/4/6. Both baseline means contain slow
samples, so neither the 22.28% nor the 15.59% reduction is a stable
kernel-speedup estimate; the minimum per-rank medians, 768.750 ->
723.700 us and 763.200 -> 731.550 us, are the steadier figure. SWA
decode measures 591.023 -> 590.482 us on 0/2/4/6, so it establishes no
material latency gain. The card-group difference is retained and no
samples or completed runs are discarded.

Separate level-4 captures on 0/2/12/14 put the HCA hca_raw_attn AIC
group span at 65.36 -> 39.98 us, hca_cmp_qk_pv at 136.21 -> 65.09 us,
and the SWA QK/PV span at 48.00 -> 39.75 us over 24 task records per
rank. Each span is one rank's first AIC task start through its last
AIC task end for that group, reported as the median across four ranks.
The raw and compressed groups overlap and the spans include gaps and
waits, so these reductions are neither additive nor comparable to the
benchmark means. Raw task records stay at 20 per rank; compressed
records go 16 -> 24 as the work becomes query-major.

Standalone a2a3 device 0 under the same 100/5 convention: drafter
B16/S7/start256 is 666.803 -> 516.348 us, generic prefill attention
plus projection at 129 tokens/ratio4 is 1096.709 -> 917.565 us, and
streaming HCA attention plus projection at 17 tokens with 0..400
compressed rows is 732.591 -> 731.444 us, which is no material change.


Little-oil:
hca
前：
image
后：
image

Little-oil:
swa
前：
image

后：
image

PR #1189, opened 2026-09-09
Perf: migrate MTP decode and prefill QK/PV pipelines

Little-oil:
- Migrate MTP decode SWA and shared prefill QK/PV to the explicit
  AIC/AIV pipeline used by Qwen PA and DSpark #1187. MTP uses
  independent attention implementations.
- Decode uses full 64-head matmuls and bounded score/probability
  transfers, with eight workers for the default B4/S2 case.
- Prefill uses 24 query workers, reuses Q across valid K blocks, and
  accumulates softmax statistics on-chip. Store one state per query and
  remove the downstream block merge. The `_staged_sparse_wave` helper
  serves both staged and physical-cache prefill paths while preserving
  main's dynamic query extents and source/merge task dependencies.
  Statistics scratch uses one row per query/head: 128 * 64 * 5 ->
  128 * 64 for each reusable staged wave, and the same factor-of-five
  reduction for physical segments.
- Add opt-in golden save/replay to both standalone entries.
- Preserve native and DeepSeek serving CI device logs in writable
  directories, enable runtime startup logging, and upload failure logs
  before cleanup. Keep complete serving pytest logs and record device
  status after acquiring the task-submit locks.

Decode is an implementation migration with no demonstrated speedup. Its
128-row SWA window fits one K block, so there is no cross-block Q reload
to remove. The benchmark windows are fully valid, leaving no invalid
block to skip. Both versions dispatch eight tasks for eight queries;
each worker processes one query, so the prelaunch cannot overlap
successive queries. The baseline already reuses KV across its two
32-head batches and performs contiguous bulk gather. Invalid-row masking
does not skip whole-block QK/PV computation. Explicit transfer and
synchronization costs remain and were not separately measured.

The measured pre-rebase ratio-0 prefill baseline already skips empty
compressed blocks. Its observed improvement includes a reduction from
128 to 24 AIC task records; it must not be attributed solely to Q reuse
or newly skipped invalid blocks. The contributions of the combined
changes have not been isolated.

Historical pre-rebase swimlane4 captures use PYPTO_BENCH=0 on the same
card. QK/PV group spans are decode 12.70 -> 13.22 us (8 -> 8 AIC
records) and prefill 139.32 -> 64.14 us (128 -> 24 records). Each span
runs from the first AIC task start to the last task end, including
setup, gaps and waits. These are single captures, not benchmark means.

Before the main CP refactor, native A2/A3 validation passed for default
and short/causal decode, partial prefill, ratio 4/128, and full SWA
callers using their existing output/cache comparators. Sparse-edge
replay covers whole-block holes, permuted pages, nonzero sinks,
transfer-slot reuse, an exactly zero empty output and inactive tails.

The first unchanged baseline launch timed out before validation or any
benchmark sample. The same compiled image passed a diagnostic replay;
subsequent fresh compilation, all timed rounds and profiling passed. The
initial failure is retained and its cause is unresolved. No measured
sample was removed.

Rebased onto origin/main 4a49d76, including the shared prefill CP
refactor in #1186. The pipeline now resides in `_staged_sparse_wave`
instead of the removed `sparse_attn_math`. Fresh native checks pass for
default decode, standalone physical prefill, the sparse-edge fixture,
and staged prefill with 259 active queries in a 264-row tensor. The last
case exercises wave starts 0/128/256, sparse holes, nonzero sinks, and
exact-zero empty queries and inactive tails. Validation uses PyPTO
afd6f60, simpler 39ce891, PTOAS 0.60, PTO ISA 5a4f74c, CANN 9.0.0, one
native A2/A3 card, PYPTO_BENCH=0 and swimlane0. Performance and full
serving have not been remeasured on the rebased source.

Before the rebase, [native A2/A3 CI on c80091c](https://github.com/hw-native-sys/pypto-lib/actions/runs/34425545280/job/102709916270)
passed. [DeepSeek serving CI on that revision](https://github.com/hw-native-sys/pypto-lib/actions/runs/34425545280/job/102709916254)
started successfully but failed exact completion-text comparison.
A subsequent controlled local repeat completed eight requests for each
version: unmodified main 52cd064 passed 7/8 and the PR sources passed
6/8. Both used the same eight cards, checkpoint, serving 559b59e and CI
execution pins; 64 prompt tokens, 128 completion tokens, temperature
0.2, top-k 32, seed 42. Every failure differed only in exact text and
had the same continuation as the CI mismatch. Thus the text mismatch
also occurs without this PR; these small samples do not establish its
root cause or exclude an effect from the PR. Test expectations and
tolerances remain unchanged.

The [A2/A3 simulator failure](https://github.com/hw-native-sys/pypto-lib/actions/runs/34425545280/job/102709916298)
is attributed to simpler and is deferred for this PR as requested.
No simulator workaround or skip is added here.

Historical pre-rebase standalone attention plus output-projection means:
decode B4/S2 139.586 -> 139.349 us (0.17%, effectively unchanged);
prefill B1/128 tokens, ratio 0, 422.801 -> 392.809 us (7.09% reduction).
Native A2/A3 device 0, one rank, PYPTO_BENCH=1, swimlane0, 100 timed
rounds after five warmups, the same passing frozen inputs/golden, and
all samples included. Means are calculated from 100 printed raw samples
at 0.1-us sample precision. Baseline is cec1268, including #1187; these
measurements predate the main prefill CP refactor and do not measure the
rebase onto 4a49d76. These are MTP standalone workloads, not full MTP
inference or the DSpark 16 x 256 TP4 benchmark.


Little-oil:
prefill
前：
image

后：
image

PR #1190, opened 2026-09-09
Perf: spread the CSA qk_pv plan over token-tile lanes

Little-oil:
- Spread csa_slots_build_valid_qk_plan across 16 SPMD lanes, each
  processing independent eight-token tiles in a strided loop, shortening
  the serial plan stage between Top-K publication and qk_pv.
- Pad each valid_block_mask row to a whole 64-byte cache line so
  concurrent scalar stores cannot overwrite another lane's rows. The
  mask/bias arithmetic and the downstream plan dependency are unchanged.

On a2a3 TP4 (local B16/S8, uniform start position 256), replaying a
frozen golden over 100 timed rounds after 5 warmups, the fastest-rank
median falls from 1052.20 to 1010.05 us (-4.01%) across nine A/B pairs,
and the per-rank plan span from 52.86-55.50 us to 7.56-9.78 us.


Little-oil:
前：
image
后：
image

PR #1199, opened 2026-09-10
Add: the Gated DeltaNet layer of Qwen3.8-27B

vloncar:
First in a series implementing Qwen3.8-27B in PyPTO, with the whole
model -- prefill and decode -- as the goal. This one adds the model
directory and the Gated DeltaNet linear-attention layer.

- Add `models/qwen3_8_27b/`. The model interleaves three GDN layers
  with one gated-attention layer over 64 layers, so 48 of them run
  this path.
- Six operators for the chunked delta rule, split as the reference
  PTO-ISA implementation splits it: `chunk_cumsum` (chunk-local prefix
  sum of the gate), `scaled_dot_kkt` (the gated key-key matrix `A`),
  `solve_tril` ((I + A)^-1 by triangular-inversion doubling),
  `wy_fast` (the WY representation), `chunk_h` (the inter-chunk state
  recurrence) and `chunk_o` (the chunk output). Each is a standalone
  entry point.
- `gdn_layer.py` composes all six as one program. `A`, `A_inv`, `W`,
  `U`, the per-chunk state snapshots and `V_new` are allocated inside it
  and never reach the caller. `build_kernel` takes `inline`, so one
  source is either a program of its own or a callee here rather than
  the operators being duplicated.
- Grouped-query attention throughout: 48 value heads against 16 QK
  heads, value head `h` reading key head `h // 3`. q and k carry Hg
  heads; v, beta, the gate, `A`, `W`, `U`, the state and the output are
  per value head. `hg` is a build parameter, so an ungrouped shape is
  still buildable and still tested.
- `config.py` carries the published `config.json` values under their
  HuggingFace names so they can be checked against it without a
  translation table, and keeps the 128-token chunk separate as what it
  is -- a tiling choice, not a model constant. Every operator takes its
  shape from there, so the default build is the model's own.
- `reference.py` is a float64 torch chain of all six stages.
  `test_gdn_stages.py` feeds each operator the reference outputs of the
  ones before it, isolating that stage; `--chain` instead feeds each the
  previous kernel's device output, which is what the deployed pipeline
  does. Both score on relative Frobenius norm, the criterion the
  reference implementation's own tests use -- an FP32 golden cannot be
  matched elementwise by kernels whose operands are FP16.
- The reference is cross-checked rather than self-consistent: it agrees
  with megagdn-pto's own GQA reference to 0.0e+00 relative on all six
  stages at (48, 16) in float64.
- `bench.py` times every stage over the reference benchmark's sequence
  lengths and writes JSON.
- Docs page and nav entry, marked under development.

Not included, and named on the docs page: the four input projections,
the depthwise kernel-4 convolution over q/k/v, the output gate, the
gated RMSNorm and `out_proj` -- the rest of the GDN block, and the next
PR in the series. The gated-attention layers, the dense MLP and the
full forward follow those. Decode is a separate operator rather than
this one at one token, since the delta rule there is the single-step
recurrence and not the chunked form; it comes after the prefill path
composes. KDA and variable-length batching are out of scope
throughout.

The target is a2a3, for now -- A5 is on the roadmap but not covered
here. Neither of CI's automated jobs reach this directory on A5: the
daily model-tests-a5 sweep is scoped to deepseek_v4_pro, and this PR's
own sim job filters its a5sim leg to examples/. What A5 evidence exists
is a manual run: all six stages and the layer pass on a5sim, same shapes
as below.

Validated on a2a3 hardware and on the a2a3 simulator. At T = 8192 all
six stages and the layer pass at (48, 16) and at (16, 16); the layer
scores 3.486e-04 relative Frobenius against the whole-chain float64
reference and runs in 3938.3 us, against 3965.0 us for the six stages
dispatched separately.

Against the hand-written PTO-ISA kernels in megagdn-pto, at the
ungrouped shape both sides currently support in our harness -- a2a3
device 0 under a task-submit grant, ours and the reference back to back
on the same card, T = 8192, H = Hg = 16, D = C = 128:

| stage | PyPTO us | megagdn-pto us | megagdn / PyPTO |
|---|---:|---:|---:|
| chunk_cumsum | 17.0 | 17.7 | 1.04x |
| scaled_dot_kkt | 112.8 | 215.2 | 1.91x |
| solve_tril | 466.3 | 512.0 | 1.10x |
| wy_fast | 129.8 | 123.4 | 0.95x |
| chunk_h | 416.6 | 420.2 | 1.01x |
| chunk_o | 320.9 | 283.3 | 0.88x |
| **total** | **1463.4** | **1571.8** | **1.07x** |

PyPTO is `effective_us`, 50 rounds after 5 warmup, median of the
per-round samples; megagdn-pto is `batched_launch_us`, 50 iterations of
20 launches, median of the per-batch samples, with library, stream and
workspace lookups hoisted out of the timed region as its own
`benchmarks/kernel/bench_gdn_kernels.py` does. The same shape at
T = 4096 reads 0.91x and at T = 16384 reads 1.16x. Both columns come
from one paired sitting and are internally comparable; the toolchain
has moved since, and our side alone now reads 1429.4 us at that shape.

The comparison at the model's own (48, 16) is not quoted because our
benchmark wrappers for megagdn-pto do not yet take a key-head count,
though its kernels support one.


vloncar:
All five addressed in 4798c15. Two notes where I did not follow the
suggestion exactly.

**`--save-output` silently turns on chaining** — real, and the worst of
the five, since it made the summary line untrue. Fixed by separating
capture from substitution: `check_stage` now takes `chain`, and
`produced` is only the capture buffer.

I kept the implication rather than removing it, though. `--save-output`
writes the original inputs beside the final output, and that pair is
only an end-to-end result if every stage ran on the previous kernel's
output. Feeding the stages the reference chain would make the saved
artifact meaningless for its stated purpose. So `--save-output` still
implies `--chain`, but now says so in its help, in the module docstring
and in the summary line, which reports `previous kernel output`.

**`chunk` must be a power of two** — real and silent, which is what
makes it worth a guard. `chunk=96` gives `ndouble=5`, so the product
stops at `A^63` while `A` is nilpotent only at `A^96`, and the kernel
returns a truncated inverse with no other symptom. Now refused at build
time.

**`run_config()` on the layer** — added, so the ring heap the layer
cannot run without travels with the kernel rather than living only in
`__main__`.

I did not make the matching change to `models/deepseek_v4_pro/`. It has
the same pattern, but it is another model's code and outside this PR;
worth its own change if a maintainer agrees.

**TOTAL row hides a shape at `0.0`** — real. `all(vals)` cannot
distinguish a stage that measured zero from one that was never timed,
and `all([])` is `True`, so a shape with no records printed a total of
`0` rather than `-`. Now checks for `None` explicitly.

**Stale GQA docstring** — real. It also pointed at a `pipeline`
function that is called `compute`; both fixed.

Revalidated after the changes: the six stages pass on `a2a3sim` in the
isolated and `--save-output` paths, the guard refuses `chunk=96` and
still builds 128, and the layer passes through `run_config()` at
3.497e-04 relative Frobenius.

zhangqi-chen:
Thanks for the thorough work on the GDN math. Before this can land, the kernels need to follow the repo's kernel form and coding style. As written, the directory reads like a separate project inside `models/`. Canonical references: `docs/pypto-coding/l2-programming.md`, `docs/pypto-coding/naming-and-comments.md`, and the `fmt-coding-style` rules in `.claude/skills/fmt-coding-style/SKILL.md`.

## 1. Remove the test and benchmark scaffolding

The kernel file's own `__main__` is the test, and `PYPTO_BENCH=1` is the benchmark. Please delete:

- `test_gdn_stages.py`: the stage orchestration, `--chain`, `--save-output`, the `OUTPUTS` / `CHAINED` maps and the capture-by-comparator trick.
- `bench.py`: the custom timing loop, the JSON output and the table. The harness already prints `[RUN] effective_us`, and the repo's metric is its median, not the mean this script reports.
- In `reference.py`, the chain machinery: `compute`, `stage_inputs`, `lazy`, `_CACHE`, `STAGES`.

What stays: each stage file has `build_tensor_specs`, `golden_<name>` and a `__main__` that calls `run(...)`, like every other model file. Each file generates its own inputs in `build_tensor_specs`. If you want to share the torch reference functions (`cumsum`, `kkt`, `solve_tril`, …) and `make_inputs`, a small helper module is fine; `glm5_3_flash/golden.py` is a precedent. Do not feed one stage from another stage's cached reference output.

`gdn_layer.py` is the end-to-end check, and its golden should compute from the tensors it receives, not from a seed-keyed cache.

## 2. Kernel form: no `build_kernel` factories

No other model builds kernels through a Python closure factory. Use the repo's single-body / two-JIT-forms idiom (see `models/deepseek_v4_flash_mtp/decode_csa.py:279-280`):

```python
def _chunk_h(
    k: pl.Tensor[[T_DYN, HG, D], pl.FP16],
    ...
    v_new: pl.Out[pl.Tensor[[T_DYN, H, D], pl.FP16]],
):
    ...

chunk_h = pl.jit.inline(_chunk_h)      # composed by gdn_layer
chunk_h_test = pl.jit(_chunk_h)        # this file's __main__ entry
```

Concretely:

- **Shapes come from module constants read off `config`**, not from `t/h/d/chunk/hg` build arguments. Drop the `--heads` / `--qk-heads` knobs. The kernel is built for Qwen3.8-27B's (48, 16), and an ungrouped build is not something the model needs.
- **Make the token axis dynamic.** Use `T_DYN = pl.dynamic("T_DYN")` in the signatures, `bind_dynamic` at the `@pl.jit` entry, and loop bounds such as `pl.spmd(t_dim // CHUNK_TILE)`. See the "dynamic B/S" section of `l2-programming.md` and `models/deepseek_v4_flash_mtp/rmsnorm.py:66-74`. Today every sequence length is a separate compile, which will not work once prefill composes. Python-side shape arithmetic has to become fixed tiles plus a documented divisibility requirement. For example, `group_tile()` in `chunk_cumsum.py` becomes a plain `GROUP_TILE`. The state tensor's leading dim depends on T, so give it its own dynamic dim, bound at the entry.
- **No build-time signature switches.** Remove `solve_tril`'s `out_dtype` parameter and pick one output dtype. FP16 is what `wy_fast` consumes, so the standalone entry should validate FP16 too.
- **No unused parameters "for a uniform signature"** (`d` in `chunk_cumsum` and `solve_tril`). Once the factories are gone, the reason for them goes too.
- Put the case size in `config.py` (a `PREFILL_SEQ`-style constant) instead of repeating `T = 8192` in seven files.

## 3. Comments: what, not why

`naming-and-comments.md`: a comment tersely states what a non-obvious line does, or there is no comment. Please strip:

- **Module docstrings down to one line** naming what the file computes. Formulas are fine. The derivations (the e^90 overflow argument in `chunk_h`, the column-scaling argument in `wy_fast`, the Neumann-series proof in `solve_tril`) belong in the PR description or the docs page, not in the kernel header.
- **All references to megagdn-pto and the reference implementation**, including in comparator docstrings.
- **History and version notes**, e.g. `chunk_o.py:71` "Fixed upstream 2026-09-02; this spelling still builds against stock pypto". If it is fixed, write the natural form (`pl.exp(g_col)`) and delete the comment.
- **Rationale prose** such as "Head-major keeps a head's chunk contiguous…", "No cast: the store carries the dtype…", "The accumulator has to be a matmul result…", and the ring-heap paragraph in `gdn_layer.py`.

The only allowed "why" comment is a pointer to a **filed** toolchain issue. Several of these comments describe compiler limitations:

- `pl.split` needing 197632 B of a 188416 B buffer (`chunk_h.py:63`);
- the `pto.tmov` address-space-pair failure under `pl.split`;
- an elementwise op after a `[1,C]->[C,1]` reshape losing `pl.split` tracking;
- `pl.create_tensor` scratch corrupting intermittently across parallel chunks (`chunk_o.py:82`).

If those are real, please file them against pypto / ptoas and reduce each comment to `# <what> (pypto#NNNN)`. If they are not reproducible any more, write the natural form.

## 4. Formatting: 110 columns, one statement per line

The files are wrapped at about 80 columns with hanging indents. The repo's `ruff.toml` target is 110, and the rule is one statement per line:

- **Do not wrap a statement's arguments.** If it does not fit, split the op into named intermediates. E.g. `chunk_o.py:98`:
  ```python
  o_acc = pl.add(inter, intra)
  o_flat[t0 : t0 + CHUNK_TILE, d0 : d0 + D] = pl.cast(o_acc, target_type=pl.FP16, mode="rint")
  ```
- Function signatures other than kernel signatures, `argparse.add_argument(...)` calls and short tuple literals such as `STAGES = (...)` go on one line.
- Kernel signatures stay one parameter per line; the `run(...)` call stays one option per line.

## 5. Header constants

Group module constants under `# model config` and `# tiling`, not `# model shape` / `# case shape`. Every tiling knob ends in `_TILE`: `CHUNK` → `CHUNK_TILE` (config.py itself calls it "our tiling choice"). `GROUP_TILE` and `COL_TILE` are already fine.

---

Most of §3–§5 is mechanical. §1 and §2 are structural and are the important part: once each stage is a `_body` + `inline` / `_test` pair on `T_DYN`, `gdn_layer.py` composes the inline forms directly, and the rest of the scaffolding has nothing left to do.

PR #1209, opened 2026-09-14
Perf: tune DSpark drafter scheduling and inverse RoPE

hashiqiqixian:
- Fan attention merge-norm over the 48 publish workers with an inner
  range instead of one SPMD block per token-head tile, and build the
  inverse RoPE from an indexed Tile gather over precomputed swap
  indices instead of the scatter/mask chains, keeping the FP32 division
  and BF16 rounding.
- Pad the drafter attention output one full row per block instead of
  512-column slices, and size the head-unpad fan to the active row
  count instead of branching over the padded maximum.
- Widen the embedding hidden tile 512 -> 1024 and narrow the shared
  expert activation block 8 -> 2 rows.
- Clear inactive gate rows by reinterpreting an FP16 zero tile as INT8
  rather than converting it, since positive FP16 zero is byte-identical.
- Drop the HC-head split-K fan, which was already K=1; the seed task,
  its dependency edge and the atomic accumulation go with it, and the
  M=16 tail becomes a single core-group scope.
- Pipeline the context-parallel payload transfer, leaving the
  notify/wait/retire protocol unchanged.

Measured on a2a3, TP4/EP8, batch 4 per rank, 100 rounds after 5 warmups
against main c286ec25 with a frozen golden replayed. Per-round maximum
across the eight ranks: median 18.307 -> 15.188 ms (-17.04%), mean
18.537 -> 15.113 ms (-18.47%). The interleaved ABBA segments disagreed
sharply (paired median reductions 29.76% and 1.61%) and the baseline
itself drifted -17.46% between them, so this is an observed win on that
host rather than a stable ratio.


hashiqiqixian:
Validation for the isolated gate change in `159e9d0`, based on `d6f2920`:

- Only `gate.py` is changed. The experimental small-batch/EP/TP settings
  from `d80e413` are not included.
- Same-card, same frozen nonzero inputs/golden, a2a3 device 3,
  256 experts, 7 active tokens out of 128, D=4096, layer 40,
  seed 20260911, 100 measured rounds after 5 warmups:
  - Raw median: **604.85 -> 97.45 us (-83.89%)**.
  - Mean: **606.041 -> 97.141 us**.
  - Each consecutive 25-round segment retains the improvement.
- Current-main candidate also passes 0/14/128 active-token cases and
  the 7-token hash-routing case. All four inactive output tails are
  checked for exact zero, in addition to the existing numerical/TopK
  comparisons. No golden or tolerance changes.
- Earlier validation of the identical gate AST covered 12 cases across
  32/256 experts and token boundaries. Current-module build and runtime
  validation passed; the final task completed with exit 0 and released
  its device lock. Full local diff review and all four hooks passed.
- Toolchain: PyPTO `8e7c98d`, simpler `4e4d3a4`, PTO ISA `a804045`,
  PTOAS 0.60, CANN 9.0.0, Python 3.10.20. Compiler and CPU thread
  pools were capped at two.

**Scope and integration caveat:** this PR establishes a standalone gate
wall-time improvement, not an end-to-end drafter/serving speedup. An older
synthetic EP8/TP4 integration comparison using the same gate change passed
golden checks but had divergent timing metrics:

| Per-rank batch | Fastest-rank median, baseline -> candidate | Per-round max-rank median, baseline -> candidate |
| --- | --- | --- |
| BS1, contemporaneous baseline | 8.46300 -> 6.88655 ms | 14.78025 -> 15.49355 ms |
| BS2, older baseline | 8.72495 -> 7.27410 ms | 14.21990 -> 20.38965 ms |

The persistent slow-rank differences are unresolved; they have not been
proven to be launch skew. These older integration results are not a
measurement of current main, which has since changed other MoE/attention
modules. No end-to-end acceptance claim is made, and this PR must not be
used as evidence that complete drafter performance is ready.

The full BS1 generated orchestration was identical apart from a consistent
one-to-one renaming of inline IDs; submission sites and explicit dependency
sites were unchanged. This does not prove identical dynamic scheduling.

No extra test files or benchmark instrumentation are included in the PR.

PR #1219, opened 2026-09-14
CI: add daily MTP and DSpark operator performance

wangqin1723-max:
- Replace broad Daily CI benchmarking with twenty MTP and DSpark
  scenario rows for CSA, HCA, SWA, MoE EP8 and projection-only LM-head
  at 8K and 128K contexts, preserving correctness and E2E coverage.
- Execute sixteen distinct workloads with five warmups and 100 raw
  rounds. Share context-independent MoE and LM-head measurements between
  report rows. Compare compatible previous and weekly runs using minimum
  rank medians of summed operator dispatch times, with rank spread shown.
- Keep TP1 Attention and TP1/DP1 LM-head at MTP B4/S2/T8 and DSpark
  B16/S8/T128 loads. Use EP8 MoE with 16 experts/rank and default routing.
  Apply the MTP expert count only in CI capture before model imports;
  keep the MoE entry and its standalone defaults unchanged.
  Add MTP TP1 projection with padded matmul rows and real-owner output,
  preserving numerical references and tolerances.
- Keep orchestration, capture and publication in ci_operator_perf.py,
  with one command parser and a separate fixed workload configuration.
  Reuse setup-ci-job's pinned environment and verify the assigned host.
  Submit two allocations: device 4 for fourteen single-card measurements,
  then eight even devices for both EP8 measurements. Hold each allocation
  across models while keeping model processes and results separate.
- Record model and group wall times with configurable case, group, queue
  and suite limits. Reserve cleanup time before each launch. Distinguish
  timeouts from device faults, clean process groups, and continue only
  after verified recovery. Preserve completed measurements across queue
  and publication errors and collect late results after queue cleanup.
- Publish performance independently from the correctness summary and
  support manual performance-only runs. Retrieve compatible history from
  dated GitHub artifacts, upload bounded evidence, and delete private
  builds only after upload. Report progress every thirty seconds and
  document device topology, measurement conventions and acceptance.
- Retry transient GitHub metadata and artifact GET failures with bounded
  backoff, including interrupted response reads, while preserving
  certificate checks, token isolation and terminal failure reporting.


tests/test_daily_ci_summary.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) PyPTO Contributors.
```
zhangqi-chen: delete all tests diff

models/deepseek_v4_flash_mtp/decode_moe.py
```diff
@@ -8,7 +8,8 @@
 # -----------------------------------------------------------------------------------------------------------
 # ci: devices=2
 """DeepSeek-V4 MoE single-layer (decode), FLASH preset. --ep picks the EP world
```
zhangqi-chen, line 10: no need to change for decode-moe

.github/scripts/ci_operator_perf.py
```diff
@@ -0,0 +1,360 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
```
zhangqi-chen, line 2: try to simplify scripts, and merge them into one.

zhangqi-chen (changes requested):

zhangqi-chen (approved):

PR #1222, opened 2026-09-15
Perf: split DSpark indexer scores across Vector lanes

Little-oil:
- Split the DSpark CSA indexer score epilogue across both Vector lanes
  with pl.split_aiv(2, UP_DOWN) and pl.aiv_shard: each lane owns 64
  candidate rows with all heads, keeping the dequantization, ReLU,
  weighting and reduction order.
- Load the query scales and head weights once per query in a shared
  full-width Vector phase, and schedule the score leaves with
  pl.cross_core_slot(slot_num=2).
- Gather two KV-scale pages per lane, clamp tail reads to a valid page
  and store disjoint 64-score ranges.
- Allow early resolution of weights_proj, indexer_topk_group_wave and
  indexer_topk_query_merge.
- Give the TP1 CSA decode a 1 GiB ring heap and forward --runtime-dir
  to both the TP1 and distributed harness paths.

Previously one lane computed the whole 128 x 64 score tile while the
other did only scale loads and control. Full CSA TP1 decode medians on
a2a3 (B16/S8/T128, 5 warmups / 100 rounds, frozen golden replay): 128K
3469.09 -> 2896.87 us (16.5%), 8K 1153.20 -> 1114.66 us (3.3%).


Little-oil:
前：
image
后：
image

PR #1223, opened 2026-09-15
Perf (V4-mtp): reduce DeepSeek V4 CP prefill state and gather copies

MaxwellF1:
- Carry only live CSA boundary state between leaves, preserving history
  and terminal state across empty leaves.
- Parallelize HCA state initialization and publish its updated view
  without a full-state copy, retaining the state-reuse dependency.
  Assign each history row to its clearing worker for exclusive writes.
- Gather contiguous HCA raw KV in 16-row tiles with boundary fallbacks,
  and use an indexed gather for the RoPE channel permutation.


MaxwellF1:
Fixed in 46bb478: each history-state row is now restored only by the initialization worker that owns and clears it, eliminating cross-worker writes. @coderabbitai review

Issue #1238, opened 2026-09-15
[Feature] Implement DeepSeek V4.1 Flash C2A Full prefill and decode


lwDavid:
### Summary

Implement the `decode_c2a_full` and `prefill_c2a_full` kernel bodies in
`models/deepseek_v4_1_flash/`, and give both files the same runnable A5
precision-and-timing entry that `decode_swa.py` / `prefill_swa.py` gained in
#1225. Both bodies are `raise NotImplementedError` today.

This is milestone 2 of #1205 ("Implement C2A Full, then validate Full-to-Reuse
cache and Top-K replay"). C2A Full is the ratio-2 KV and index-key **source**
for backbone layers 2, 8 and 14, so nothing downstream of it — C2A Reuse, the
40-layer prefill/decode loop — can be validated until it lands.

The work also requires one correction to the cache ABI that #1207 scaffolded;
see "Packed-FP4 payload ABI" below. That correction is the reason this issue is
filed before the implementation rather than after.

### Motivation / Use Case

`models/deepseek_v4_1_flash/` currently has one of six attention modes
implemented. SWA (layers 0-1) is done and validated on A5. C2A Full is the next
mode by dependency order:

- It is the only writer of the ratio-2 compressed KV pool and of the ratio-2
  index-key pool, both addressed by the shared `c2a_cmp_kv` block table.
- It owns the ratio-2 recurrent compressor state, the only cross-call carrier of
  a half-finished token pair.
- It publishes `topk_indices`, which the C2A Reuse layers (3-7, 9-13, 15-19)
  replay verbatim without re-running the indexer.

Every other ratio-2 mode consumes state that only this mode produces, so C2A
Full is on the critical path for the first end-to-end text milestone in #1205.

#### Kernel ABI

The 48-parameter `@pl.jit.inline` signatures established by #1207 are kept in
parameter name and order, with the single dtype/shape correction below. The two
files stay identical except for `output_window`'s capacity
(`DECODE_MAX_TOKENS` vs `PREFILL_MAX_TOKENS`).

#### Packed-FP4 payload ABI (correction to #1207)

`pl.FP4` reserves **one byte per logical element** on device, not a nibble.
Measured on A5 (Ascend950DT, ptoas v0.61): `pl.FP4.get_bit()` is 4 but
`pl.FP4.get_byte()` is 1, and a `pl.Tensor[[8, 64], pl.FP4]` round-trips 64
bytes per row with row 1 starting at byte offset 64.

[...]

PR #1241, opened 2026-09-15
Add: implement DeepSeek V4.1 Flash C2A full attention

lwDavid:
- Implement the decode and prefill kernel bodies on one shared ratio-2
  pipeline: compressor projection and recurrent pool, compressed KV and
  index-key publication, indexer scoring, device Top-K, and sparse
  attention over the sliding window plus the selected compressed rows.
- Declare the compressed KV and index payloads as packed UINT8 at half
  the logical width. pl.FP4 reserves one byte per logical element on
  device, so the scaffold shape did not match the packed E2M1 layout
  that quantize_mxfp4_cache emits and the cache budget assumes.
- Select every reachable position when a token can see at most
  index_topk of them, and otherwise fold leaf-sorted candidates into a
  running Top-K so the sort width stays independent of the context.
- Carry physical cache rows through the sort, which drops a per-entry
  block-table lookup from the selection result.
- Pool a completed pair under a per-channel two-way gate softmax, and
  leave each request's last even-position row in the recurrent state so
  a pair split across two calls resolves against the carried row.
- Split the FP32 compressor weights into two BF16 terms so the Cube
  reproduces the reference product from a BF16 activation.
- Add an independent CPU reference transcribed from the official
  implementation that rewrites only addressed cache rows, decode and
  prefill fixtures, per-output comparators, and an A5 precision and
  timing entry.

Refs #1238
Refs #1205


lwDavid:
## Validation report

### Source of truth

DeepSeek-V4.1-Flash `inference/model.py` and `inference/kernel.py` at revision
`dba1be0a40aa45a94ad051997016db3960a90277` — `Compressor`, `Indexer`, `Attention`,
`sparse_attn`, `fp4_act_quant`. The in-repo `official_reference_c2a` is a transcription of
that math, not an execution of the CUDA kernels.

### Scope

`models/deepseek_v4_1_flash/decode_c2a_full.py` and `prefill_c2a_full.py` only. Both kernel
bodies, their shared ratio-2 pipeline, the independent CPU reference, fixtures, comparators and
the A5 precision/timing entry. The 48-parameter ABI from #1207 is unchanged in name and order
except for the two packed-payload declarations described below.

Not in scope: C2A Reuse, any C1A mode, the layer-20 candidate selector, mHC, EP-MoE, the
40-layer loop, and the whole-model peak-HBM measurement #1205 asks for.

### ABI correction

`pl.FP4` reserves **one byte per logical element** on device. Measured on A5:
`pl.FP4.get_bit()` is 4 but `get_byte()` is 1, and a `pl.Tensor[[8, 64], pl.FP4]` round-trips
64 bytes per row with row 1 starting at byte offset 64. The scaffolded
`[.., 1, HEAD_DIM] pl.FP4` therefore reserved 512 B/row while `quantize_mxfp4_cache` emits
256 B/row, and the per-card cache budget in #1205 assumes the packed form. The two C2A files
now declare `[.., 1, HEAD_DIM // 2] pl.UINT8` and `[.., 1, INDEX_DIM // 2] pl.UINT8`, matching
`models/deepseek_v4_pro/expert_routed.py`. Scale tensors are unchanged. The same two
declarations still need correcting in the four C1A and two C2A Reuse scaffolds; they are left
alone here to avoid conflicting with #1235.

### Environment

| Component | Version |
|---|---|
| pypto-lib | `fcf978b` (branch `feat/deepseek-v4-1-flash-c2a-full`) |
| pypto | `1deae896` |
| simpler | `39ce891d` |
| pto-isa | `5a4f74cb` |
| ptoas | 0.61 |
| torch | 2.10.0+cpu |
| Device | Ascend950DT, 8 cards, openEuler 24.03 LTS-SP4 aarch64 |

### Acceptance thresholds

Output: global rel-L2 <= 1%, every row <= 2%, cosine >= 0.9999, exact zero rows — the budget
#1216 established. Recurrent state: rel-L2 <= 0.5%. Cache ownership: bytes outside the mapped
slots must be bit-identical. Cache payloads: published rows within 1% relative L2 of the
reference decoded values. Top-K: same set of physical rows, no duplicates, worst-row overlap
>= 0.98. These are engineering acceptance thresholds, not published DeepSeek tolerances.

### Device results, TP4/DP1, 4 cards

| Case | decode rel-L2 | decode cosine | prefill rel-L2 | prefill cosine |
|---|---|---|---|---|
| mixed | 0.00335853 | 0.99999195 | 0.00670476 | 0.99996179 |
| long | 0.00681908 | 0.99997592 | 0.00737004 | 0.99996459 |
| masked | 0 | - | 0 | - |
| zero | 0 | - | 0 | - |

All eight pass. `mixed` keeps every token below `index_topk` reachable positions, so the
selection takes everything; `long` pushes part of the batch past `index_topk` and exercises the
scored Top-K. `masked` clears every slot and index; `zero` zeroes activations and all three
cache payloads. Worst row across the eight: 0.0150945.

TP1/DP1, 1 card, `mixed`: decode rel-L2 0.00227474 cosine 0.99999440, prefill rel-L2 0.00464922
cosine 0.99997187.

Every case also checked, per rank: all six cache tensors byte-exact on published rows with
untouched bytes bit-identical; `compressor_state` rel-L2 1.1e-6; `topk_indices` worst-row set
overlap 1.000000. At TP4 the four TP replicas of `output` are byte-identical.

### Reproduction

```
python models/deepseek_v4_1_flash/decode_c2a_full.py -p a5 --tp 4 --dp 1 -d 0,1,2,3
python models/deepseek_v4_1_flash/prefill_c2a_full.py -p a5 --tp 4 --dp 1 -d 0,1,2,3
```

Add `--case {mixed,long,masked,zero}`, `--tp {1,2,4}`, `--tokens`, `--requests`, `--seed`.
`PYPTO_BENCH=1` with `--epochs N` runs the timing path.

### Timing

Functional verification of the timing entry, not a claimed speedup. Decode, TP4, 24 tokens,
6 requests, 4 epochs per dispatch, 100 rounds, all four ranks resident: fastest-rank median
11158.8 us, fastest-rank min 10247.6 us for the four-epoch dispatch. Compilation, input
generation, the CPU reference and host total runtime are excluded. The host is shared, so the
mean and max are inflated by co-tenants; the min and median are the usable figures. No tuning
was attempted and no comparison baseline exists.

Unlike SWA this operator is not idempotent across epochs — the ratio-2 compressor carries
request-scoped state, so a replayed epoch reads what the previous one published. The reference
is replayed the same number of times, so a benchmark dispatch stays accuracy checked
(rel-L2 0.00354906 at four epochs).

### Known differences

- Comparing the device against `attention_common.golden_compressed_attention` instead of the
  transcription gives rel-L2 0.0163-0.0264. That gap is the reference, not the kernel: the two
  existing SWA references disagree with each other by rel-L2 0.0168168 / cosine 0.99985814 on
  the same inputs. Two arithmetic choices account for it — `_project_output` runs the grouped
  `wo_a` einsum as a plain BF16 torch einsum while the A5 kernel accumulates in FP32, and
  `merge_attention_stats` keeps FP32 softmax probabilities while the device casts them to BF16
  before the PV matmul. Setting both to the golden's convention in the transcription and
  changing nothing else drops the disagreement to 0.000148-0.000936 with 100% Top-K overlap and
  bit-identical cache values.
- `wo_b` is an MXFP8 GEMM with dynamic E4M3 activation quantization, which is a step function of
  the latent. Measured amplification: a latent perturbation of 0.17% becomes 0.71% at the
  output, 0.47% becomes 1.67%, 1.0% becomes 2.69%. Any reference whose latent differs from the
  kernel's by more than one BF16 ULP therefore looks 1-2% off.
- The E2M1 grid carries eight magnitudes, so a one-ULP difference in a published row can land on
  a table midpoint and move a code one step; the device fuses the rotary multiply-add that torch
  rounds twice, which is enough to do that. Observed rate across the seeds exercised: at most
  one byte in 3072 published payload bytes, and in that instance the single flipped nibble moved
  the output rel-L2 from 0.002 to 0.0118 through the amplifier above. Cache ownership stays a
  byte-level contract; published payload codes are compared as decoded values.
- Top-K row sets can differ at the selection boundary because `relu` leaves many candidate
  scores exactly equal. The comparator is set-based and tolerates that; in every case reported
  here the overlap was in fact 1.000000.

### Not verified

- TP2, DP2, and EP8. Only TP1 and the production TP4 at DP1 were run.
- Context lengths beyond the fixtures used here. The longest `compressed_lens` exercised is
  4096 compressed positions per token; the scored path was reached but not at the 1M-token
  scale #1205 targets.
- Full-to-Reuse cache replay and prefill-to-decode continuity across separate calls. Both are
  C2A Reuse dependencies and neither mode exists yet.
- Peak HBM per card, and any end-to-end generation.
- Performance. No tuning was attempted; the numbers above only show the timing entry works.

PR #1247, opened 2026-09-16
Add: DeepSeek V4.1 mhc

meromber2-eng:
- Implement the complete mHC kernel for DeepSeek V4.1 Flash, including coefficient generation, pre collapse, post residual expansion, and head collapse.
- Add FP32 RMS normalization, tiled split-K linear projection, sigmoid gating, and Sinkhorn normalization for residual mixing, with support for dynamic token counts and the existing dtype ABI.
- Add end-to-end golden validation with deterministic inputs, output and dtype checks, and configurable platform, device, batch, sequence, and compile-only options.


lwDavid:
Reviewed against the bar the merged V4.1-Flash operators set. The math is right:
all four kernels match `golden.py`'s `hc_mixes`/`hc_pre`/`hc_post`/`hc_head`
operation for operation, including the Sinkhorn prologue and the
`residual_mix[t, in_h, out_h]` contraction, and it passes on A5 from 1 to 4096
tokens including every partial-tile tail. `# ci: a5` is correct and does select
this file on current main.

Five things to fix before merge.

## 1. The fixture saturates every sigmoid, so the self-test is blind

`build_mhc_tensor_specs` draws `function` from an unscaled
`torch.randn(MIX_HC, HC_DIM)` with `HC_DIM = 20480`, so the projection has
std ~145 and every sigmoid and softmax in `hc_mixes` is hard-saturated:

```
pre  takes 4 distinct values   {1e-06, 0.0006, 0.9963, 1.0000009}
post takes 3 distinct values   {1.4e-21, 7.8e-06, 2.0}
inv_rms = 0.998 / 1.0018       (i.e. ~1)
```

Scoring mutated references with this PR's own fixture and its own tolerances:

| mutation | pre_mix | post_mix | residual_mix |
|---|---|---|---|
| **drop `inverse_rms` entirely** | PASS | PASS | PASS |
| **`scale[1]` doubled** | PASS | PASS | PASS |
| **`scale[2]` doubled** | PASS | PASS | PASS |
| `scale[0]` doubled (control) | FAIL | PASS | PASS |

So the inverse-RMS normalization and two of the three `scale` entries are not
covered at all — the operator could lose them and stay green. Scaling
`function` by `1/sqrt(HC_DIM)` puts the sigmoids back in their active range and
restores the coverage.

## 2. `sys.path` is rewritten at import time

Lines 17-21 run at module scope with no guard: they remove the repository root
from `sys.path` and re-insert it at position 0. That happens whenever the
network imports this module, not only when it is run as a script, and it demotes
the importing program's own `sys.path[0]` (measured: repo root moves from index
1 to 0). Every merged sibling guards it:

```python
if __package__ in (None, ""):
    sys.path.insert(0, str(Path(__file__).resolve().parents[2]))
```

## 3. No `main()`

The whole harness is inlined into the `__main__` block. Every merged sibling
defines `main()` and has the sentinel call it, which keeps the validation path
importable and the entry block to three lines.

## 4. No per-tensor precision is printed

A passing run prints only PASS/FAIL. The merged operators print a `[PRECISION]`
line per tensor with the achieved error, which is what makes a green run
reviewable and a regression visible. The numbers are good here — rel-L2 1.2e-5
to 3.0e-5 on the coefficient tensors and 1.3e-4 to 2.8e-4 on the BF16 ones, with
max deltas at one BF16 ULP — but the entry does not report them.

## 5. Drop "draft" from the title

The repository squash-merges with the PR title, so "Add: DeepSeek V4.1 mhc
draft" is what lands on main. The sibling subjects read like "Add: implement
DeepSeek V4.1 Flash C2A full attention". The branch commit subject also starts
"Addi:".

---

Every number above is reproducible and I can share the scripts. Happy to be
wrong on any of it.

lwDavid:
Re-checked the new head on A5. The four harness items are in, and the operator
passes at 1, 2, 7, 8, 9, 16, 17, 31, 128, 1000 and 4096 tokens.

One correction to my own review: scaling `function` was the wrong lever for the
inverse-RMS gap, so that one is still open. `inverse_rms` is
`rsqrt(mean(x_hc**2))` — it depends on `x_hc`, not on `function`. `init_x_hc`
draws standard normals over `HC_DIM = 20480`, so the mean square concentrates at
1.000 ± 0.002 and removing the normalization from `hc_mixes` entirely moves the
coefficients by at most 0.07%. Scored with this file's own tolerances:

| mutation | pre_mix | post_mix | residual_mix |
|---|---|---|---|
| drop `inverse_rms` | PASS | PASS | PASS |
| `scale[0]` doubled (control) | FAIL | PASS | PASS |
| `scale[1]` doubled | PASS | FAIL | PASS |
| `scale[2]` doubled | PASS | PASS | FAIL |

The rescale did buy the last two rows — both were PASS/PASS/PASS before it.

Giving the token rows different magnitudes closes the first one:

```python
def init_x_hc():
    magnitudes = (0.5 + torch.arange(tokens) % 4).reshape(-1, 1, 1)
    return torch.randn(tokens, HC_MULT, D, generator=generator) * magnitudes
```

With that line, dropping `inverse_rms` fails all three (4.8% / 10.3% / 19.8% at
the default two tokens, and it grows with token count), no other row in the
table changes, and the kernel still passes on A5 at 2, 8, 9, 128 and 4096
tokens. Nothing else from my original five is outstanding.

lwDavid (approved):
Approved.

PR #1249, opened 2026-09-16
Perf: stack pages and fuse DSpark CSA indexer Top-K

Little-oil:
- Assign score work per (query, leaf) item with a stride of 24 workers
  instead of having every long-history worker walk all queries.
- Stack eight cache pages into 256-row INT32 score matmuls, pipeline
  the score loop with two stages, and use one cross-core transfer slot
  so the larger vector fragment fits within UB.
- Fold the positive query dequantization scale into the head weight
  once per query/leaf instead of multiplying every score batch twice.
- Score contiguous half-leaves and select their Top-512 inside the
  score task (512/1024/2048/4096-row sorts), then merge the compact
  pair rows sequentially; this removes the group-sort task stage.
- Reuse score storage across leaves: score/pair buffers total 36 MiB
  instead of 136.375 MiB at TP1, batch=16, S=8.
- Keep direct Top-K publishing for batches whose longest cache fits in
  one 8192-candidate leaf.
- Round the maximum leaf count up so a partial final leaf has reserved
  pair rows, matching the runtime leaf calculation.

Full DSpark CSA mean device effective_us against b1a456d6 on the same
a2a3 NPU (TP1, batch=16, S=8, frozen inputs, 600 timed samples per
case): 8K 940.504 -> 921.995 us (-1.97%), 128K 2694.501 -> 2243.626 us
(-16.73%). Page stacking, direct item assignment, coefficient folding
and fused leaf selection are adapted from #1236. Goldens and
comparison tolerances are unchanged.


Little-oil:
Addressed the `TOPK_MAX_LEAVES` sizing comment in `075b0d4`. The maximum segment count now uses ceiling division, matching the runtime calculation, so a partial final segment has reserved pair rows. The current model's count remains 32.

Validation: a2a3sim with a model limit of 8193 compressed candidates passes scores, Top-K indices, and both cache outputs; all pre-commit checks pass. The Summary and source links are updated.

The previous `sim (a2a3sim)` CI failure is the already tracked FFTS simulator compatibility issue in CSA's `qk_pv` kernels ([failed job](https://github.com/hw-native-sys/pypto-lib/actions/runs/35071464414/job/104713805888)). The standalone indexer passed in that same simulator job, and native A2/A3 CI passed. No simulator tests or synchronization were disabled.

Little-oil:
@coderabbitai Fixed in 075b0d4: TOPK_MAX_LEAVES now uses ceiling division. The 8193-candidate boundary case passes validation.

PR #1254, opened 2026-09-16
Add: Implement and validate DeepSeek V4.1 Flash decode C1A on A5

superwjckawaii:
- Implement the DeepSeek-V4.1-Flash decode C1A family on A5: decode_c1a_full.py (operator plus shared kernels and the validation harness), decode_c1a_reindex.py, decode_c1a_reuse.py and hierarchical_sparse_indexer.py.
- Share projection, RMSNorm, interleaved RoPE, paged low-bit gather, combined window plus compressed online attention with one sink, inverse RoPE, grouped output projection and the TP reduction across the three decode operators. FULL owns compressor/index cache publication and candidate generation, REINDEX scores against the byte candidate mask, REUSE consumes caller-supplied top-k rows.
- Accumulate scale-corrected group-32 MXFP8 products without expanding the whole weight, and publish persistent caches as explicit packed UINT8 carriers with hand-packed E2M1 codes and E8M0/E4M3 scales; keep the candidate mask INT8 because A5 has no boolean output.
- Resolve E2M1 midpoints half-to-even, matching the publish kernels' compare chain, and floor E8M0 cache scales at 2^-126.
- Validate with a TP1/TP2/TP4 host that pages the window, compressed and index caches, compares every cache, scale, top-k and mask byte-exactly, reports output relL2 under the formal allclose tolerances, and ships a CPU --check-fp4 midpoint check.
- Implement the hierarchical sparse indexer with tiled block maxima, a fixed scalar min-heap in UB, causal-tail pinning and logical-order tie selection, emitting the candidate blocks as an INT8 mask.


high-cloud:
**[P1] The validation harness aliases pure outputs with their golden references, producing false PASS results.**

In [`run_c1a`, lines 1830–1832](https://github.com/hw-native-sys/pypto-lib/blob/e63d80566af32effe52b7abb5b90e5dd187b0efd/models/deepseek_v4_1_flash/decode_c1a_full.py#L1830), every argument receives `TensorSpec(..., init_value=values[n])`, including pure outputs. `TensorSpec.create_tensor()` returns `self.init_value.to(dtype=self.dtype)`, which returns the same tensor when the dtype already matches.

`_prepare_inputs()` calls `create_tensor()` for the device buffers; `_compute_golden()` calls it again for pure outputs. Consequently, `output`, `topk_indices`, and FULL's `candidate_mask` share storage with their expected results. Device readback overwrites the golden, and validation compares the result against itself. The `InOut` caches follow a separate clone path and are not affected in the same way.

Minimal CPU demonstration:

```python
value = torch.zeros(1)
spec = TensorSpec("output", [1], torch.float32, init_value=value)
actual = spec.create_tensor()
expected = spec.create_tensor()
assert actual.data_ptr() == expected.data_ptr()
actual.fill_(42)
assert expected.item() == 42  # The reference was overwritten too.
```

**Hardware evidence:** on revision `995e94f`, the original FULL/REINDEX/REUSE TP1 commands all reported PASS with output relL2=0. A diagnostic wrapper that only makes each `TensorSpec.create_tensor()` return a clone, without changing kernels or golden math, exposes these failures:

| Mode | TP1 output relL2 | TP2 output relL2 | Failed tensors |
|---|---:|---:|---|
| FULL | 0.766625 | 0.769124 | `topk_indices`, `output` |
| REINDEX | 0.730140 | 0.719440 | `topk_indices`, `output` |
| REUSE | 0.574929 | 0.608861 | `output` |

All tested caches/scales pass byte-exact comparison; outputs fail the existing `rtol=atol=1e-3` check. Cases use the defaults: tokens=1, pages=8, epochs=1, seed=17, case=random.

Environment: Ascend950DT/A5, CANN 9.2.0, PyPTO `fb107da9`, its pinned simpler `22385d2b`, PTOAS 0.61, PTO ISA `3b4faf67`, torch 2.14.0+cpu. This is a newer, internally pin-matched stack than the PR's CI environment. Numerical failures have not been localized to kernel vs. reference vs. compiler. I rechecked the current head `e63d805`: the initializer/aliasing pattern remains; the hardware measurements above are from `995e94f`.

Please ensure independent output/reference storage before relying on these PASS results. A local fix is to omit tensor initializers for pure outputs or use factories that return a fresh clone on every call (cloning once when constructing the spec is insufficient). Add a regression that deliberately changes the actual output and verifies that validation fails, then rerun numerical validation with independent buffers.

superwjckawaii:
> **[P1] The validation harness aliases pure outputs with their golden references, producing false PASS results.**
> 
> In [`run_c1a`, lines 1830–1832](https://github.com/hw-native-sys/pypto-lib/blob/e63d80566af32effe52b7abb5b90e5dd187b0efd/models/deepseek_v4_1_flash/decode_c1a_full.py#L1830), every argument receives `TensorSpec(..., init_value=values[n])`, including pure outputs. `TensorSpec.create_tensor()` returns `self.init_value.to(dtype=self.dtype)`, which returns the same tensor when the dtype already matches.
> 
> `_prepare_inputs()` calls `create_tensor()` for the device buffers; `_compute_golden()` calls it again for pure outputs. Consequently, `output`, `topk_indices`, and FULL's `candidate_mask` share storage with their expected results. Device readback overwrites the golden, and validation compares the result against itself. The `InOut` caches follow a separate clone path and are not affected in the same way.
> 
> Minimal CPU demonstration:
> 
> ```python
> value = torch.zeros(1)
> spec = TensorSpec("output", [1], torch.float32, init_value=value)
> actual = spec.create_tensor()
> expected = spec.create_tensor()
> assert actual.data_ptr() == expected.data_ptr()
> actual.fill_(42)
> assert expected.item() == 42  # The reference was overwritten too.
> ```
> 
> **Hardware evidence:** on revision `995e94f`, the original FULL/REINDEX/REUSE TP1 commands all reported PASS with output relL2=0. A diagnostic wrapper that only makes each `TensorSpec.create_tensor()` return a clone, without changing kernels or golden math, exposes these failures:
> 
> Mode	TP1 output relL2	TP2 output relL2	Failed tensors
> FULL	0.766625	0.769124	`topk_indices`, `output`
> REINDEX	0.730140	0.719440	`topk_indices`, `output`
> REUSE	0.574929	0.608861	`output`
> All tested caches/scales pass byte-exact comparison; outputs fail the existing `rtol=atol=1e-3` check. Cases use the defaults: tokens=1, pages=8, epochs=1, seed=17, case=random.
> 
> Environment: Ascend950DT/A5, CANN 9.2.0, PyPTO `fb107da9`, its pinned simpler `22385d2b`, PTOAS 0.61, PTO ISA `3b4faf67`, torch 2.14.0+cpu. This is a newer, internally pin-matched stack than the PR's CI environment. Numerical failures have not been localized to kernel vs. reference vs. compiler. I rechecked the current head `e63d805`: the initializer/aliasing pattern remains; the hardware measurements above are from `995e94f`.
> 
> Please ensure independent output/reference storage before relying on these PASS results. A local fix is to omit tensor initializers for pure outputs or use factories that return a fresh clone on every call (cloning once when constructing the spec is insufficient). Add a regression that deliberately changes the actual output and verifies that validation fails, then rerun numerical validation with independent buffers.
**The aliasing is real and fixed.** `TensorSpec.create_tensor()` returned `init_value` itself when the dtype matched, so `output`, `topk_indices` and FULL's `candidate_mask` were compared against themselves. Pure outputs are now built through per-call clone factories, and the regression you asked for is in: a deliberately corrupted actual output now fails validation (it used to pass).

**With independent buffers the failures reproduce** (our A5: FULL 0.7565, REINDEX 0.7267, REUSE 0.5813, same `topk_indices` mismatches), exposing two real bugs, both fixed:

1. **Tile mask scatter drops the even lane on A5.** `gather_combined` and `decode_index_keys` unpacked the FP4 nibbles with `pl.tile.scatter_mask(..., P0101/P1010)`, which silently zeroes half of every compressed and index key on A5. The unpack now uses the tensor mask scatter like every other kernel in the repo.
2. **Device-only scratch buffers had no dependency edges.** A task whose operands create no tensor-map edge is ready at submit, so the index scorer could overlap its producers (decoded keys / quantized query / projected weights) and read them half-written — the TP2/TP4 flaky where the byte-exact top-k flipped between two stable states (6/8 failures before). Every producer now returns its `pl.Scalar[pl.TASK_ID]` and every consumer carries it in `deps=[...]` from prepare through index, score, top-k, gather, attention and the TP reduce. This holds 10/10 consecutive TP4 reindex runs now.

**Re-validated on head `f582bb9`:** TP1/TP2/TP4 × {full, reindex, reuse} pass (output rel_l2 ≤ 0.31% against the 1% FP8 budget; all caches, scales, candidate masks and top-k rows byte-exact), TP4 reindex 10/10, ragged / masked / sink / tokens=1 pass, 16512-wide mask passes, compile-only 9/9, CPU `--check-fp4` 14/14. Note your pinned stack is newer than CI's, so absolute rel_l2 values may shift slightly.

Thanks again — the original "all PASS with relL2=0" was a vacuous comparison that was hiding two real correctness bugs.

PR #1259, opened 2026-09-17
Fix: keep DeepSeek prefill within the scalar ABI limit

MaxwellF1:
- Read the CSA history destination bound from its tensor shape inside the
  core scope instead of capturing an additional scalar argument.


MaxwellF1:
Reduced this PR to the one-line scalar ABI fix. Cache/state correctness changes moved to draft #1279, paired with hw-native-sys/pypto-serving#238; they are not required by this ABI-only PR.

Validation: all pre-commit hooks and CP8 code generation passed on base 68d9c584. The identical one-line patch on frozen base 2602cbb6 passed the eight-device cold and prefix scenarios (39 cold requests, 13 prefix requests). TTFT changes versus main were -0.7%, -0.8%, and +5.6% at 64, 1202, and 8192 tokens; one sequential run cannot establish whether the long-input difference is caused by this edit. Prefix-hit median TTFT was 2.983 seconds.

Scope: remove the scalar-argument overflow that causes INVALID_ARGS; this does not resolve the separate cache persistence semantics or prove that earlier intermittent timeouts are fixed.

PR #1279, opened 2026-09-18
Fix: preserve DeepSeek prefill cache rows and boundary state

MaxwellF1:
- Persist every current raw KV row and reusable CSA/HCA boundary state.
- Return all current pre-HC rows through the dynamic prefill output.
- Size slot metadata for all CP segments and preserve cache-write ordering.
- Extend the existing golden checks for cache rows and boundary states.


MaxwellF1:
Split from #1259. This draft contains only cache/state correctness changes; the scalar ABI fix remains in #1259 and is a prerequisite. Pair with hw-native-sys/pypto-serving#238.

The six changed files plus the separate ABI patch exactly match the full-fix files used in the completed eight-device comparison. On that frozen baseline, full-fix TTFT versus ABI-only increased 6.6% / 17.4% / 31.2% for 64 / 1202 / 8192-token cold prefill, and 7.2% for a 1152-token prefix hit in a 1202-token prompt. These are one sequential run, three warmups and ten measured requests per case; they are not new-main measurements.

Current split: all pre-commit hooks and 13 serving contract tests passed. The paired run produced ten tokens on every request; 1202-token outputs passed the existing CI text check. This does not establish full numerical correctness or eliminate the earlier intermittent tensor_wait_timeout. Keep this PR in draft while reviewing correctness coverage, timeout stability and performance costs.

PR #1280, opened 2026-09-18
Fix: Guard V4.1 paged access and gather serving-owned RoPE rows

### LINKED ISSUE #1278: [Bug] V4.1 paged metadata, RoPE table integration and C2A page reads
## Background

V4.1 needs correct paged-cache metadata, graph-side access to caller-owned full RoPE tables, and visibility checks before C2A indexer page reads. Implementation: PR #1280.

## Problems and expected behavior

| Area | Problem | Expected behavior |
| --- | --- | --- |
| Paged metadata | Incomplete compression groups and window padding can trigger unnecessary page lookup; required pages may be missing. | Skip unpublished/padding rows and reject missing required write/history pages. |
| RoPE composition | Attention needs token-major rows from caller-owned full tables; gathering inside every layer repeats work. | Gather once per forward for each profile/positions/active-count combination, then reuse rows across matching consumers. |
| C2A indexer | Shared Full prefill/decode `index_select` reads page-table/cache entries before checking tile visibility. | Skip completely invisible tiles before lookup and fill invalid outputs; preserve scoring, physical rows, Top-K and publication dependencies. |

Before the C2A fix, `length=1` with a one-column table reaches `p0=128` and reads nonexistent column 1. `length=513` with five columns reaches `p0=640` and reads nonexistent column 5. Length zero also enters the fixed scan. These are source/CPU address findings, not reproduced NPU crashes. Partially valid 64-row tiles remain within allocated 128-row pages and retain their mask.

## Implementation and validation

PR #1280 implements the three areas above and now links this issue with `Fixes #1278`. The PR is still open; this issue remains open until merge. Current reviewed head: `07d0df5645a46dd7d6a7283e60d33eb6f656c842` (four commits).

- Current-head pre-commit, docs, unit-tests and `a5-v41-flash` CI passed. [A5 device job](https://github.com/hw-native-sys/pypto-lib/actions/runs/35700517073/job/106657225613). The decode composition CI includes mixed requests at TP=1/4; this is not evidence for the complete boundary matrix below.
- Earlier SSH validation at `ae0d2b1` (production files matched `02c2bd8`) passed 18 A5 compilation configurations and six physical Ascend950DT cases across C2A decode, attention prefill and mHC prefill, including mixed requests and two epochs. Generated orchestration retained fixed RoPE widths and gathered rows before repeated attention. Stack: PyPTO `e8191e3`, PTOAS 0.61, CANN 9.2.0. These results predate the latest rebase.
- Earlier 24 metadata host tests and 10 exact A3 RoPE helper cases passed.
- Earlier A5 compilation and nine exact A5 simulator indexer cases passed at `c8c8208`: lengths 0/1/128/129/512/513/2048/2049 plus mixed requests, minimal tables and non-contiguous pages. Generated reads were guarded and `publish_ready` retained. These are historical simulator results, not physical-device validation of the current revision.

The previously recorded PyPTO `c9af905` composition compilation blocker is not a current blocker on the validated newer stack.

Validation tracking:

- [x] Compile and numerically validate composed RoPE on a compatible A5 stack, including generated dependency inspection (historical SSH validation; current-head A5 CI also passes).
- [ ] Run the complete C2A boundary matrix above on physical A5 hardware. Current CI success does not establish this coverage.

Validation code remains on independent branches, outside PR #1280:

- https://github.com/hashiqiqixian/pypto-lib/tree/validation/pr1280-rope-reuse-rebased-20260921
- https://github.com/hashiqiqixian/pypto-lib/tree/validation/pr1280-pages-rope-20260920

No measured latency, full-model or M0 acceptance is claimed. External serving adapter migration is separate. Compressor ring state is #1260 / PR #1298; serving loading, shared-page copy-on-write and TP-to-EP are outside this issue.

hashiqiqixian:
- Skip page lookup for incomplete compression groups and window padding;
  reject missing required write and visible-history pages.
- Gather query and compressed RoPE rows from caller-owned full tables
  before repeated C2A attention, retaining token-major layer interfaces
  and producer dependencies. Reuse rows for matching profiles, positions
  and active token counts; use identity rotation for negative positions.
- Guard invisible C2A indexer tiles before page-table and cache reads in
  both short and long histories. Fill invalid outputs while preserving
  valid scoring, physical row IDs, Top-K and publication dependencies.
- Document full-table ownership and row reuse, remove the unused host
  selector, and retain row-based inputs in the decode composition
  adapter.

Fixes #1278


high-cloud:
Two review points on `6c0d40d`:

1. **[P2] Fix the metadata field name in the integration documentation.** In `docs/models/deepseek_v4_1_flash/index.md` (lines 61 and 84), `metadata.compressed_rope_positions[source]` should be `metadata.compressed_rope_position_ids[source]`. `ForwardMetadata` exposes only the latter, so following the documented integration raises `AttributeError`. The attention argument name `compressed_rope_positions` can remain unchanged.

2. **Avoid redundant RoPE gathering when composing multiple attention layers.** Accepting serving-owned full tables is useful, but it does not require gathering the same rows inside every attention invocation. Both new Full wrappers currently allocate four row tensors and invoke `materialize_rope_rows` twice per call; each helper uses `pl.spmd(32)`. When layers share the same profile, positions and active token count, this repeats row reads/writes and introduces additional scheduling/dependency work. This is a performance concern, not a measured latency regression.

   The existing decode forward paths provide useful precedents: `deepseek_v4_flash_mtp/decode_fwd.py` calls `gather_decode_rope_rows` once near the start and reuses the resulting rows across layers; `deepseek_v4_flash_dspark/decode_fwd.py` accepts token-major rows prepared by the caller and forwards them to the layers. Some standalone MTP prefill entries gather internally, so helper availability alone does not establish the intended full-model composition.

   Please make the intended composition explicit: prepare rows once per forward for each distinct profile/position combination, then reuse them through the row-based attention interface. The retained `c2a_full_partial` already allows this; full-table wrappers can remain convenient standalone entries. If repeated gathering is intended for the production multilayer path, please provide an end-to-end latency comparison, particularly for small-batch decode.

Validation note: the configured `a5-v41-flash` PR job selects these entries and runs real-device golden validation with TP1/DP1 and the default single-epoch mixed case. Passing that job can cover default numerical validation; it does not establish the latency cost of repeated gathers or cover repeated epochs/TP4.

PR #1281, opened 2026-09-18
Perf: tune DSpark HCA TP1 cache waits and normalization

wangqin1723-max:
- Pass separate RAW and compressed cache-write task IDs through the
  TP1 HCA attention helper. Remove their shared dummy so each gather
  waits for its corresponding cache writer, preserving inferred tensor
  dependencies and the standalone sparse-attention entry.
- Enable the existing reciprocal-row normalization path for TP1 HCA.
  comb_sinkhorn computes each row-sum reciprocal before broadcast
  multiplication in the initial softmax and iterative normalization.

Full HCA passes the existing compress_state, kv_cache, cmp_kv and x_out
comparisons with both changes. The normalization changes only the
generated comb_sinkhorn kernel; the dependency split changes
orchestration while all 34 generated kernel bodies remain identical.

On main 856bc9f without #1265, one same-allocation comparison on physical
device 14 measures main at 708.1395 us, the dependency split d570d9d at
721.9895 us, and the split plus Sinkhorn normalization at 710.9710 us.
The new commit lowers the original PR median by 1.53%; the final PR is
still 0.40% slower than main in this comparison. These are individual
100-sample comparisons, not repeated stability measurements. 8K and TP4
were not benchmarked.

The earlier dependency-only experiment used 6902ee8 from #1265, which
includes RAW KV reuse in L1: median 828.8805 -> 754.2600 us (-9.00%) on
physical device 15. That conditional result is separate from the
current-main comparison above.

All timings use a2a3, TP1, B16, S=8, context 131072 per request,
PyPTO f1bb086, simpler 15f5cbd, PTOAS 0.57, PTO ISA 96ba706 and CANN
9.0.0. Each variant runs in one process with the same frozen golden,
5 warmups and 100 samples, retaining every sample with chip swimlane,
dep-gen, PMU, dump-args and scope-stats disabled. The incremental
full-HCA effective_us median is 721.9895 -> 710.9710 us (-1.53%).


wangqin1723-max:
Consolidated the selected TP1 HCA changes into #1265 as two separate
commits: RAW KV reuse in L1, followed by the cache dependency split from
d570d9d. The combined PR records the profiling-off 8K and 128K results.
The reciprocal Sinkhorn commit 657fd42 is not included; this branch is
preserved so that change remains available separately. Closing this PR
as superseded by #1265.

PR #1289, opened 2026-09-18
Refactor: modularize V4.1 decode attention composition

bumble0918:
- Support hidden-width RMSNorm inputs used by decode compositions
- Split SWA and C2A Full/Reuse leaves from their mHC compositions
- Add rank-aware host entries and shared composition contracts
- Add layer planning and golden dispatch for six attention variants
- Share projection, RMSNorm, RoPE, Q/KV, and output projection stages
- Document decode composition boundaries and readiness

Refs #1205 and #1271 (tasks 1 and 3; task 2 remains out of scope).


tests/contract/test_deepseek_v4_1_flash_contract.py
```diff
@@ -0,0 +1,396 @@
+# Copyright (c) PyPTO Contributors.
```
high-cloud, line 1: 这个文件删了吧

bumble0918: > 这个文件删了吧

暂时保留吧，加了一些公共计算提取的check，避免重复实现

high-cloud:
**[P2] Make the C1A native ABI guard effective before dispatching**

In [decode_layer.py, lines 171–175](https://github.com/hw-native-sys/pypto-lib/blob/905a2503f277176e825312be31428bbb45bb3eb6/models/deepseek_v4_1_flash/decode_layer.py#L171-L175), the guard checks `COMPOSITION_ABI == "native"`, but none of the three selected C1A modules defines that attribute. They therefore default to `"spec-driven"`, and `mode.make_program(world_size, epochs, specs)` passes the arguments to a factory whose actual signature is `make_program(tokens, pages, epochs=1)`.

CPU-only reproduction at the current head:

```python
from models.deepseek_v4_1_flash.decode_layer import make_decode_layer_program

program = make_decode_layer_program(20, 4, 2, specs=[])
```

Instead of raising the intended `NotImplementedError`, this returns a C1A leaf host with `x.shape == [4, 4, 5120]` and captured `EPOCHS == []`: the world size became the token count, the epoch count became the page count, and the specs list became the loop bound. Layers 24 and 21 reproduce the same misdispatch.

Please explicitly mark the selected C1A modules as native ABI (or encode that distinction in the dispatcher), so they are rejected before JIT construction, and add rejection-path tests for all three C1A modes. The existing 42 contract tests pass but do not exercise this guard.

PR #1294, opened 2026-09-18
Add: chain DeepSeek V4.1 prefill C2A attention with mHC

lwDavid:
## Summary

Run one V4.1 attention sublayer the way the official `Block.forward`
does: `hc_mixes` on the incoming residual stream, `hc_pre` with the
previous sublayer's delayed pre-mix, `attn_norm`, packed prefill C2A
(Full or Reuse) with its TP reduction, then `hc_post`. The pre-mix this
sublayer produces comes back as `next_pre_mix`, which is what the FFN
sublayer of the same layer collapses its input with.

The files follow the layout the SWA chain uses: the leaf attention
operators move to `prefill_attn_c2a_full.py` and
`prefill_attn_c2a_reuse.py`, while `prefill_c2a_full.py` and
`prefill_c2a_reuse.py` hold the mHC-wrapped sublayers, the new
`attn_norm` kernel and their shared validation. All four remain A5
entries, so the leaf and the chain are guarded separately.

`attn_norm` is new: it is the per-layer RMSNorm over the hidden size
that sits between `hc_pre` and attention. At D=5120 a row does not fit
in UB whole, so it reduces and applies in 8x512 tiles; the existing
`make_norm` helpers keep a row resident and serve widths 512 and 1280.

## Why the validation is stage by stage

The attention operators are discontinuous in their input: flipping 13 of
245760 elements of the attention input by one BF16 ULP moves the
golden's own output by up to 0.89% (four trials, CPU). The device's
`hc_pre` and RMSNorm reduce in a different order than torch and flip
about that many, so comparing the chained output against a pure golden
gates the change on the MXFP8 group scale and the E2M1 grid instead of
on the wiring. Fed the chain's golden input, the unmodified
`prefill_c2a_full` returns 0.395%.

Each stage is therefore checked against the stage in front of it:

- `attn_input` against the golden collapse and norm at `hc_pre`'s
  budget, with every token row inside 2**-8, one BF16 ULP everywhere in
  a row.
- attention output, caches, compressor state and Top-K against a
  reference re-run on the device's own `attn_input`, at the unchanged
  C2A budgets and the same byte-ownership contract.
- `x_hc_out` against `hc_post` replayed on the device's own attention
  output, plus an end-to-end bound.
- `next_pre_mix` at `hc_mixes`' budget.

End-to-end differences are printed and not gated; at TP1 they are 1.0%
for the attention output and 0.25% for the residual stream.

The fixture gives each token a different magnitude, so neither RMS
statistic is a no-op. Fifteen hand-written golden mutants - skipped
norm, norm without its weight, this sublayer's mix instead of the
delayed one, transposed comb, halved post, attention fed a buffer other
than `attn_input`, one wrong token row, and others - are all rejected by
these comparators in both modes at 48 and 256 tokens, and the unmodified
golden passes.

## Test plan

On A5 with the stack the job pins (pypto 1deae896, ptoas 0.61, torch
2.10), run serially, both entries:

- the two chained entries at TP1, TP2, TP4 x DP1, DP2, that is 1 to 8
  cards, on the default case
- the two chained entries with `--case long`, `--case masked`, `--case
  zero`, `--tokens 37` for the ragged tail, `--tokens 256` and
  `--epochs 2`
- the two leaf entries at TP1 and TP4, which also serve as the control
  arm: they reproduce the CI log's precision numbers on this tree
- `pytest tests` 388 passed, pre-commit clean

Device cost of the mHC stages, `effective_us` over 20 rounds, minimum,
TP1: Full 7478 -> 7730 us at 48 tokens and 36645 -> 37020 us at 256
tokens; Reuse is unchanged within run-to-run spread.

## Notes
- The SWA chain in `prefill_swa.py` collapses with the mix the same
  sublayer produces and runs no attention RMSNorm. This one follows
  `Block.forward` and the delayed pre-mix the model document describes,
  which is also what the final HC collapse uses.
- Not from this change: at 256 tokens `prefill_c2a_full` alone can
  exceed its 2% per-row output budget (standalone seed 18 reaches
  0.0224), and one run hit an empty `ptoas compilation failed` for
  `swa_mx_projection_2` that three reruns did not reproduce.


lwDavid:
## Results

Device: A5 on the host the a5 runner uses, with the stack the job pins
(pypto 1deae896, pto-isa 5a4f74cb, ptoas 0.61, torch 2.10). Every case
ran serially; the tree is this branch rebased on 2257c82. Control arm on
the same tree: unmodified `prefill_c2a_full --tp 1` reproduces the CI
log's numbers bit for bit (output rel_l2 0.00464922), so the environment
matches the job.

### Cases

Both entries, 24 runs, 24 pass:

| Case | Cards | Full | Reuse |
|---|---|---|---|
| TP1 x DP1 (what CI runs today) | 1 | pass | pass |
| TP1 x DP2 | 2 | pass | pass |
| TP2 x DP1 | 2 | pass | pass |
| TP2 x DP2 | 4 | pass | pass |
| TP4 x DP1 | 4 | pass | pass |
| TP4 x DP2 | 8 | pass | pass |
| `--case long` / `masked` / `zero` | 1 | pass | pass |
| `--tokens 37` (ragged tail) | 1 | pass | pass |
| `--tokens 256` | 1 | pass | pass |
| `--epochs 2` | 1 | pass | pass |

### Precision, TP1, 48 tokens, default case

Full:

```
attn_input      rel_l2=3.62e-05  max_row_rel_l2=1.22e-04
next_pre_mix    rel_l2=6.75e-07  max_row_rel_l2=2.51e-06
attention       rel_l2=0.00393   cosine=0.9999746  max_row=0.0130
window_cache    byte_mismatch=0/24576      compressed_cache 0/6144
index_cache     byte_mismatch=1/1536       value_rel_l2=0.0044
compressor_state rel_l2=1.09e-06           topk overlap=1.000000
x_hc_out replay rel_l2=2.74e-05  max_row_rel_l2=1.39e-04
x_hc_out e2e    rel_l2=0.00255   (printed, not gated: 0.0102 for the
                                  attention output)
```

Reuse: attention rel_l2 0.00350, cosine 0.9999772, window cache
0/24576 bytes, `x_hc_out` replay 3.14e-05.

TP4 x DP2, Full: attention rel_l2 0.00439, cosine 0.9999727, window
cache 3/24576 bytes at value rel_l2 3.96e-04, Top-K overlap 1.000000,
`attn_input` and `next_pre_mix` identical across every TP replica.

### Fixture strength

Fifteen golden mutants, each one step of the chain rewritten, scored
with this harness's own fixture and comparators, in both modes at 48 and
256 tokens: all fifteen are rejected, the unmodified golden passes. The
mutants that only the per-token-row bounds catch are the single wrong
token row ones; the ones that only `attn_input` catches are the norm and
pre-mix mistakes, since the attention stage is teacher forced.

### Timing

`effective_us`, 20 rounds, minimum, TP1, device time only:

| | attention alone | with mHC | delta |
|---|---|---|---|
| Full, 48 tokens | 7478.5 us | 7729.7 us | +251 us |
| Full, 256 tokens | 36645.0 us | 37019.9 us | +375 us |
| Reuse, 48 tokens | 6700.4 us | 6691.1 us | within spread |

### Observed, not from this change

- At 256 tokens the unmodified `prefill_c2a_full` can miss its 2%
  per-row output budget by itself: seeds 17/18/19/20 give max row
  0.0176 / 0.0224 / 0.0125 / 0.0126, so seed 18 fails standalone while
  the global rel_l2 stays under 1%. CI only runs 48 tokens.
- One run failed with an empty `ptoas compilation failed:` for
  `swa_mx_projection_2`; the same case compiled before and after, and
  three reruns produced bit-identical passing numbers.
- On pypto HEAD (after the call-argument layout check) the C2A entries,
  the ones here included, stop at `pipeline_input` with `parameter
  'wq_a_scale' is declared MX_B_NN but the argument is ND`, because the
  host orchestrator shards stacked ND scales into MX_B_NN rank
  parameters. The a5 job pins pypto, so CI does not see it.

lwDavid:
Rebuilt on the SWA layout and re-validated.

**Layout.** The leaf attention operators moved to
`prefill_attn_c2a_full.py` and `prefill_attn_c2a_reuse.py`, keeping their
own `main()` and `# ci: a5` marker, and `prefill_c2a_full.py` /
`prefill_c2a_reuse.py` now hold the mHC-wrapped sublayers. Four entries,
four separate precision guards, the same split `prefill_attn_swa.py` and
`prefill_swa.py` use. The operator and golden names follow: the leaf is
`prefill_attn_c2a_full` / `golden_prefill_attn_c2a_full`, the chain is
`prefill_c2a_full`, and the shared runner is `run_prefill_c2a`. The model
document's file table and milestone list are updated.

**AI review.** CodeRabbit produced no inline comments on this PR; the
review thread count is zero, so there is nothing to resolve. Its one
failing pre-merge check was docstring coverage at 66.67%; every function
in the four files now carries one, so coverage is 100% in the files this
PR touches.

**Device, after the rename.** 28 runs, 28 pass, serially, on the stack
the a5 job pins:

- the two leaf entries at TP1 and TP4, which double as the control arm
- the two chained entries at TP1, TP2, TP4 x DP1, DP2, that is 1 to 8
  cards
- the two chained entries with `--case long|masked|zero`, `--tokens 37`,
  `--tokens 256` and `--epochs 2`
- `pytest tests` 388 passed, pre-commit clean

Numbers are unchanged by the rename, for example Full at TP1 and 48
tokens: `attn_input` rel_l2 3.62e-05 with the worst token row at
1.22e-04, attention rel_l2 0.00393 at cosine 0.9999746, window and
compressed caches byte identical, `x_hc_out` replay 2.74e-05.

PR #1306, opened 2026-09-20
Add: DeepSeek V4.1 engram kernel with tensor-parallel support

### LINKED ISSUE #1304: [Feature] Add DeepSeek V4.1 Flash Engram kernel with tensor-parallel support
### Summary

Add a DeepSeek V4.1 Flash Engram module (gated n-gram hash memory) to
`models/deepseek_v4_1_flash/`, with optional tensor-parallel (`--tp {1,2,4,8}`)
support that row-shards the embedding table across ranks and merges per-rank
partial lookups through a distributed window all-reduce.

### Motivation / Use Case

DeepSeek V4.1 introduces an Engram module that, per token, gathers 24 rows
(8 heads × 3 n-grams) from a hash-indexed table, projects them through `wkv`
into key/value, applies a sigmoid gate computed from `q·k` RMS-normalized
dot products, and injects the gated value back as a residual. Without a
pypto-lib implementation, V4.1 pipelines cannot run end-to-end on Ascend,
and the TP behaviour of the embedding-table shard has no reference.

The existing `models/deepseek_v4_1_flash/` directory covers attention, MoE,
and HC pre/post, but lacks the Engram kernel. A single-rank kernel alone is
insufficient because production deployments run the table under TP and we
need a verifiable all-reduce path to validate equivalence.

### Proposed API / Behavior

Single-rank entry matches the existing per-kernel test pattern:

```python
python models/deepseek_v4_1_flash/engram.py -p a5 --tp 1 --device 0 \
    --batch 2 --sequence 4
```

Tensor-parallel entry adds `--tp` and a comma-separated `--device` list:

```python
task-submit --ignore-whitelist --device auto --device-num 8 --run \
  'python models/deepseek_v4_1_flash/engram.py -p a5 --tp 8 \
       --device $TASK_DEVICE --batch 2 --sequence 4'
```

Key implementation pieces:

- `engram` / `engram_gate` — single-rank kernel; gate factored out for reuse.
- `engram_tp` — TP kernel: masked per-rank gather → `pld.tensor.put` to a
  shared window → `pld.system.notify`/`wait` barrier → blocked `matmul` that
  all-reduces the partial lookups from every peer before the projection.
- `engram_tp_group` — `@pl.jit.host` orchestrator that allocates the window
  and signal buffers and launches one rank per device.
- Verified equivalent to the single-rank path: rel_l2 ≈ 3.17e-11,
  max_abs ≈ 1.05e-08 across TP ∈ {1, 2, 4, 8} on A5 hardware.

### Alternatives Considered

- **Replicated table per rank** — wastes HBM (`NUM_EMBEDDINGS × HEAD_DIM`
  BF16 per rank) and does not reflect production TP deployments.
- **`pld.system.allreduce` on the gathered rows** — couples communication
  with the matmul K-loop less cleanly than the explicit window+notify/wait
  pattern used in `decode_o_proj.py` / `attention_tp.py`; the chosen design
  matches the existing in-repo TP protocol.

### Additional Context

- Reference PyTorch implementation:
  `DeepSeek-V4.1-Flash/inference/model.py` (open-source release).
- Validation: `batch=2, sequence=4` PASS on TP = 1/2/4/8 (rel_l2 = 3.17e-11).
- Files proposed: `models/deepseek_v4_1_flash/engram.py` (~680 lines,
  self-contained: kernel + host orchestrator + golden harness + CLI).

supurs:
- Implement the gated n-gram hash memory module on A5: per token gathers
  24 rows (8 heads x 3 n-grams) from a hash-indexed table, projects them
  through wkv into key/value, applies a sigmoid gate from the
  RMS-normalized q.k dot product, and injects the gated value as a
  residual
- Fold the released module's q_weight and k_weight into a single FP32
  per-channel scale host-side so the kernel takes one weight tensor and
  the gate path avoids a BF16 rounding
- Factor the gate into engram_gate so single-rank and TP paths share it
- Shard the engram table by NUM_EMBEDDINGS // TP_SIZE rows per rank and
  mask the per-token gather so each rank only contributes the rows it
  owns; publish per-rank partial lookups to a shared window, barrier
  with pld.system.notify/wait, then all-reduce the partials inside the
  K-loop of the projection matmul
- Bound tail token tiles when t_dim is not a multiple of T_TILE: guard
  the per-row hash_ids read with t0 + tt < t_dim and zero-fill the
  out-of-range rows, then write the matmul accumulator back via
  pl.set_validshape so only valid_rows rows land in kv / lookup_partial,
  matching the pattern engram_gate already uses for x / out
- Add --tp {1,2,4,8} CLI and an engram_tp_group host orchestrator that
  allocates the window/signal buffers and launches one rank per device

Fixes #1304


supurs:
Hi maintainers, could you please add the `run-ci` label to trigger the PR test workflow? Thanks!

supurs:
Addressed the CodeRabbit actionable comment in 10c5aa2.

**Fix**: bound tail token tiles when `t_dim % T_TILE != 0`:
- Compute `valid_rows = min(T_TILE, t_dim - t0)` per token block
- Guard per-row `pl.read(hash_ids, ...)` with `t0 + tt < t_dim`, zero-fill out-of-range rows so the matmul sees a fully defined tile
- Write back via `pl.store + pl.set_validshape` (TP path) and subscript-assign + `set_validshape` (single-rank path), matching the pattern `engram_gate` already uses for `x` / `out`

**Verified on a5** with `--batch 2 --sequence 4` (t_dim=8 < T_TILE=16, hits the partial-tail path):

| TP | rel_l2 | max_abs |
|---|---|---|
| 1 | 5.26e-09 | 1.91e-06 |
| 2 | 5.26e-09 | 1.91e-06 |
| 4 | 5.26e-09 | 1.91e-06 |
| 8 | 5.26e-09 | 1.91e-06 |

Precision unchanged vs the previous commit (still bit-identical across TP), and the previously unbounded reads/writes past `t_dim` are now guarded.

PR #1308, opened 2026-09-20
Add: mHC composition to V4.1 Flash MoE

Bretty082a1:
- Compose delayed mHC pre-mixing and residual expansion around the
  existing V4.1 Flash MoE core.
- Extend the host ABI with mHC inputs and expose next_pre_mix, x_mixed,
  and the FP32 x_next residual stream for validation.
- Add deterministic mHC fixtures and routed MXFP4 weights at the
  requested decoded magnitude, with composed golden references.
- Validate complete owner and non-owner output partitions and require
  exactly one owner per active token.


lwDavid:
## Review

Read the change and ran it on A5. Environment: 2 cards, EP2 / TP1, 16
tokens, `pypto e8191e3c` / `ptoas 0.61` / `torch 2.10.0`, every arm
submitted serially so the arms cannot manufacture each other's
failures. `_moe_core` checks out: an `ast.unparse` diff of its body
against the base `moe` is empty and both take the same 32 parameters,
so the kernel itself is untouched and the review surface is the
wrapper, the fixture and the comparator.

### Device arms

| Arm | What | Result |
|---|---|---|
| A | this branch as-is | pass (`diff_thd=5e-3, pct_thd=0.05`) |
| B | this branch, thresholds back to the file's previous `3e-3 / 0.02`, no `max_abs_diff` | **pass** |
| C | this branch with `expert_routed.gen_routed_mx_weights` instead of the new fixture generator | fail, and only on `max_abs_diff` -- `worst absolute diff=64 exceeds max_abs_diff=0.25`; the ratio gate itself passed |
| D | this branch without `manual_dep=True` | pass |
| E | error-distribution probe on `x_next` | see below |
| F | base commit, unmodified `l3_moe` | pass |

Arm E, active partition, 655,360 points:

```
frac(rdiff > 1e-3) = 5.77e-03      max_abs           = 6.25e-02
frac(rdiff > 3e-3) = 4.42e-03      mean|expected|    = 9.62e-01
frac(rdiff > 5e-3) = 3.46e-03
```

The probe uses a plain `max(|a|, |e|)` denominator with no floor, so
those fractions are upper bounds on what `ratio_reldiff` itself would
count. The inactive partition is empty, because `--num-tokens`
defaults to `MOE_TOKENS`.

---

### 1. The composition uses a non-delayed `pre_mix`

```python
mhc_mixes(x_hc, hc_ffn_fn, ..., pre_mix, post_mix, residual_mix)
mhc_pre(x_hc, pre_mix, x_mixed)     # consumes the pre_mix it just made
```

`prefill_c2a_full.attention_hc_pre` and `prefill_c2a_reuse` take
`pre_mix` as an **input** and publish this sublayer's mixes as
`next_pre_mix`; `docs/models/deepseek_v4_1_flash/index.md` states the
final collapse "applies the last layer's delayed `pre_mix`", and
`golden.hc_head` repeats it. Under that contract the MoE sublayer has
to collapse with the pre-mix the attention sublayer produced, and
publish its own for the next layer's attention -- so in a real layer
the attention block's `next_pre_mix` would be dropped on the floor
here, and `moe` would need `pre_mix` in and `next_pre_mix` out.

`prefill_swa` does compute and consume in one step, so both shapes
exist in the tree today. Worth settling which contract this is meant
to satisfy before it lands.

### 2. The thresholds are looser than before, and need not be

Before: owner rows under `ratio_reldiff(diff_thd=3e-3, pct_thd=0.02)`,
every other row under `torch.equal`. Now: both partitions under
`5e-3 / 0.05`, plus an absolute `max_abs_diff=0.25`.

Arm B passes at `3e-3 / 0.02`, and arm E measures a 0.35% bad-point
ratio against the new 5% budget -- 14x headroom, and 4.5x against the
old 2%. `max_abs_diff=0.25` sits at 26% of `mean|x_next|` (0.962),
and `mhc_post` rounds its result to BF16, so 0.25 is about 32 BF16 ulp
at unit magnitude; measured worst absolute difference is 0.0625.
`docs/run-and-validate/golden-harness.md`: "Choose tolerances from the
numerical contract of the kernel, not simply to make a failing result
pass."

Separately, the TP ownership invariant is gone. Rows the rank does not
own used to have to match bit for bit; they are now inside the same
5%-of-elements budget as everything else, so a kernel that writes a
row or two it does not own can still pass.

### 3. The routed fixture magnitude moved 40x, undisclosed, in a fork

Measured decoded routed weight std: **0.997 -> 0.0246** (ratios
0.0247 / 0.0246 / 0.0244 for w1 / w3 / w2). Owner-row sublayer
composition:

| fixture | sublayer mean abs | routed part | shared part |
|---|---|---|---|
| `gen_routed_mx_weights` | 1530.7 | ~100% | 0.80 |
| this branch | 1.44 | 1.18 (82%) | 0.80 |

The change looks right: `gen_routed_mx_weights` accepts `dequant_std`
and never uses it while `gen_mxfp8_weight_kn_v41` honours its own, the
comment above the routed loop already claims "Real routed shards at
the deployment magnitudes", and with `swiglu_limit = 10.0` and
`x @ w1` at std ~72 the old routed SwiGLU was essentially all clamp.

What I would change is how: this copies forty lines of a shared helper
into `moe.py`, so the routed-expert entry keeps the saturated fixture
and the two disagree from now on; nothing in the description says the
magnitudes moved; and it rides in the same commit as the threshold
change, so a reader cannot tell which change the new numbers are for.
Fixing `gen_routed_mx_weights` itself (or giving it the parameter it
already advertises) keeps one generator. `_routed_generator=None` is
dead either way -- the only caller always passes the new one.

### 4. `_token_partition_mhc_compare` re-implements `ratio_reldiff`

`ratio_reldiff` already ships `valid_rows` / `valid_axis` ("for packed
buffers whose inactive tail would otherwise dilute the error ratio"),
`zero_tail` ("catches a kernel writing past the active token count")
and `max_diff_hd`, a single-point cap that is relative rather than a
hard-coded absolute constant. Arm C is what the absolute form costs:
it fires on fixture scale rather than on error. One
`ratio_reldiff` call with `max_diff_hd`, `valid_rows=active` and
`valid_axis=1` covers the whole comparator. Its three keyword
parameters also have no caller.

### 5. The mHC fixture drops two properties the sibling keeps

`prefill_c2a_full.make_hc_inputs` gives each token its own magnitude
and rounds `x_hc` through BF16, with the reasons in line: the mixes
divide by a per-token `inv_rms`, and the residual stream holds BF16
values. A plain `torch.randn(EP, T, HC_MULT, D)` gives every token the
same RMS -- measured std of the per-token RMS across tokens is
**0.0061** -- which collapses `scale*(mixes_raw*inv_rms)` to a constant
factor and makes any per-token indexing mistake in the RMS stage
invisible.

`hc_ffn_fn * 0.0635`, `hc_ffn_scale = [0.11334, 0.035901, 0.058183]`
and the 24 `hc_ffn_base` constants have no stated provenance; the
sibling uses `randn / sqrt(HC_DIM)`, which is 0.00699. If these came
from a checkpoint, a comment saying so would help.

### 6. Validation reach goes down, and nothing in CI runs this entry

The description notes no standalone MoE path is exposed any more, and
none of `x_mixed`, `sublayer` or the three mix tensors is an output
either. `prefill_c2a_full` exposes `attn_input`, `attn_output` and
`next_pre_mix` "so each stage can be checked" and carries a
teacher-forced reference, for a reason written into its docstring: one
BF16 ulp at the handoff moves an MXFP8 group scale and the result by
close to 1%. Without a handoff tensor a failure here is not
attributable to a stage.

Also, `moe.py` carries `# ci: devices=2`, but `detect_changes.py`
excludes `models/deepseek_v4_1_flash/` from the A2/A3 and simulator
sweeps and selects A5 entries by `# ci: a5`, so neither job picks this
file up. That predates this change, but it matters more once `l3_moe`
is the only validation path. Tagging it is not quite enough on its
own: `a5_parallel_cases.py` builds the matrix from `--tp` and `--dp`
choices, and this entry declares `--ep` and `--tp`, so the generated
card counts would be wrong and EP4 / EP8 would never run. The reported
validation is one configuration.

### 7. `manual_dep=True` does the opposite of what the comment says

`create_tensor(manual_dep=True)` opts the tensor out of OverlapMap
auto-dep tracking for its entire lifetime -- it removes edges, it does
not extend a lifetime -- and no compensating `deps=[...]` is added.
The ordering survives only transitively, through `pre_mix` ->
`mhc_pre` -> `x_mixed` -> `_moe_core` -> `sublayer` -> `mhc_post`;
the direct edge through `residual_mix` does not help, because
`residual_mix` is written by the `mhc_sinkhorn` loop rather than by
the `mhc_split` loop that writes `post_mix`. `residual_mix` crosses
the same `pl.scope()` boundary without the flag. Arm D passes without
it.

### 8. `l3_moe` is ~110 diff lines for three logical changes

Normalising both sides with `ast.unparse` against this branch's own
base leaves 3 changed statements out of 29: the signature, the
docstring, and `moe(` -> `moe_test(`. The rest is rewrapping lines
that were already inside `ruff.toml`'s `line-length = 110`, and the
same goes for the `alloc_window_buffer` / `pld.window` block. Reverting
the rewrap would cut the diff roughly in half.

### 9. The `sys.path` bootstrap is out of scope

It runs at import time for a case CI never hits -- every job exports
`PYTHONPATH` at the repository root -- no other entry in the directory
has it, and it inserts the root at `sys.path[0]` while the `__main__`
block below deliberately removes the model directory. Two path
policies in one file.

### 10. Smaller things

- No linked issue. #1275 asks for the MoE ABI to stop carrying
  attention-TP semantics and for mHC pre/post to run on the token
  shard; this wires mHC around the current `token_owners` / `tp_rank`
  ABI, so linking the two would keep them from colliding.
- `_golden_smoke.run_moe_golden` is written against the deleted
  `golden_moe(x, ...)` signature. It had no caller before this change
  either, but it can never be called now -- worth deleting or moving to
  the `golden_moe(tensors)` shape.
- `ruff check`, `check_headers.py` and `check_english_only.py` are all
  clean on the branch.

---

### Suggested minimal changes

1. Settle the `pre_mix` contract first; if it is the delayed one, the
   signature needs `pre_mix` in and `next_pre_mix` out.
2. Put the thresholds back to `3e-3 / 0.02` (arm B) and express the
   comparator as `ratio_reldiff(..., max_diff_hd=..., valid_rows=active,
   valid_axis=1)`; restore an ownership check, e.g. that non-owner
   active rows equal the pure residual expansion.
3. Fix the routed magnitude in `gen_routed_mx_weights` rather than a
   private copy, and say in the description that the magnitudes moved.
4. Drop `manual_dep=True` (arm D).
5. Give `x_hc` per-token magnitudes and a BF16 round-trip; document
   where the `hc_ffn_*` constants come from.
6. Expose one handoff tensor (`sublayer` or `x_mixed`) for stage-wise
   acceptance, and check what `a5_parallel_cases.py` produces for this
   entry before tagging it.
7. Revert the reformatting in `l3_moe` and the window block, and drop
   the `sys.path` bootstrap and the `_routed_generator=None` branch.

Happy to re-run any arm, or the same matrix at EP4 / EP8, once the
`pre_mix` question is settled.

high-cloud:
Reviewed commit `0f166b95541570482bb79db1a316009e1d8da7e2`. Two findings in the output comparator, assuming the current token-owner contract is retained:

1. **[P1] Comparing only a prefix of the masked non-owner/inactive partition misses incorrect outputs.**

   Location: [`moe.py:974–982`](https://github.com/hw-native-sys/pypto-lib/blob/0f166b95541570482bb79db1a316009e1d8da7e2/models/deepseek_v4_1_flash/moe.py#L974-L982), used at lines 1002–1005.

   `actual[other_mask].unsqueeze(0)` has shape `[1, number_of_selected_rows, HC_MULT, D]`. Passing `valid_rows=active, valid_axis=1` to `other_compare` checks only the first `active` selected rows, although this partition contains `EP_SIZE * MOE_TOKENS - active` rows. The remaining rows are silently excluded, including from NaN/Inf checks.

   CPU reproductions using the comparator and `golden/validation.py` from this commit:
   - EP4/EP8, 16 active tokens: set the last rank's non-owner token-0 row to NaN; comparison returns PASS.
   - EP2, 8 active tokens: set the last rank's inactive token-15 row to NaN; comparison returns PASS.
   - EP2, 0 active tokens: corrupt the output with NaN; comparison returns PASS because both comparisons select zero rows.

   EP2 with all 16 tokens active happens to have exactly 16 non-owner rows, so that configuration does not expose the truncation. Remove `valid_rows`/`valid_axis` from the comparators once the masks have selected their partitions, and compare each complete selected partition.

2. **[P2] The ownership check validates the total count, not exactly one owner per token.**

   Location: [`moe.py:994–997`](https://github.com/hw-native-sys/pypto-lib/blob/0f166b95541570482bb79db1a316009e1d8da7e2/models/deepseek_v4_1_flash/moe.py#L994-L997).

   `owner_active.sum() == active` allows a duplicate owner for one token to cancel a missing owner for another. In an EP2 CPU reproduction, assigning token 0 to both ranks and token 1 to neither rank still passes this check and the comparator when actual equals expected. The golden also consumes the supplied owner map, so output agreement alone does not establish this invariant.

   Validate each token independently, for example:

   ```python
   if not (owner_active.sum(dim=0) == 1).all().item():
       return False, "token_owners must select one rank per active token"
   ```

Validation scope: CPU-only comparator reproductions; no NPU execution. These findings concern false-positive validation, not a demonstrated numerical failure in the MoE kernel. They are separate from the token-sharded ABI redesign tracked in #1275.

lwDavid:
## `valid_rows` disables most of the new ownership check

Re-ran the revision on A5: EP2 / TP1 and EP4 / TP1 both pass, all three
outputs. The delayed `pre_mix`, the restored budget and ownership
split, the `ratio_reldiff` options, the fixture statistics and the
`manual_dep` removal all look right to me.

One thing came in with the new comparator, though. In
`_token_owner_mhc_compare` both gates share one budget:

```python
budget = dict(diff_thd=3e-3, pct_thd=0.02, max_diff_hd=1.0,
              valid_rows=active, valid_axis=1)
owner_compare = ratio_reldiff(**budget)
other_compare = ratio_reldiff(**budget)
```

`other_compare` is handed `actual[other_mask].unsqueeze(0)`, whose
axis-1 length is `active * (EP - 1) + EP * (T - active)`, not `active`.
`_valid_prefix` narrows to the leading `active` entries and drops the
rest, and `zero_tail` is off, so the dropped rows are not checked at
all. For `owner_compare` the same argument is a no-op, since that
tensor's axis 1 is already `active` long.

Calling the comparator directly with one whole row overwritten by
`1e4` (20480 elements), everything else equal:

| Config | Non-owner rows | Compared | row 0 | row `active` | last row |
|---|---|---|---|---|---|
| EP2, 16 tokens | 16 | 16 | fail | -- | fail |
| EP2, 8 tokens | 24 | 8 | fail | **pass** | **pass** |
| EP4, 16 tokens | 48 | 16 | fail | **pass** | **pass** |
| EP8, 16 tokens | 112 | 16 | fail | **pass** | **pass** |

`--ep` accepts 2, 4 and 8 and `--num-tokens` can be below `MOE_TOKENS`.
EP2 with the full token count is the one shape where the non-owner
group happens to be exactly `active` rows long, so it is fully covered
-- and that is the configuration the reported run uses, which is why
this does not show up as a failure.

Dropping `valid_rows` and `valid_axis` from the shared `budget` fixes
it: the two boolean masks already select exactly the rows each gate
should see, so there is no padded tail left to exclude.

```python
import torch
import models.deepseek_v4_1_flash.moe as M     # imported under --ep 4 --tp 1

owners = M._owner_pattern().unsqueeze(0).expand(M.EP_SIZE, -1).contiguous()
cmp = M._token_owner_mhc_compare(M.MOE_TOKENS)
expected = torch.randn(M.EP_SIZE, M.MOE_TOKENS, M.HC_MULT, M.D)
actual = expected.clone()
actual[1, 6] = 1e4                                   # a non-owner row
print(cmp(actual, expected, inputs={"token_owners": owners},
          actual_outputs={}, expected_outputs={}, rtol=1e-3, atol=1e-3))
# -> (True, '')
```

PR #1310, opened 2026-09-20
CI: add the V4.1-Flash simulator cases to the A5 daily

lwDavid:
The daily a5sim leg swept V4-Pro only, so the V4.1-Flash files a
simulator can run had no daily run anywhere: the A2/A3 sweeps stopped
taking that directory, and the A5 device sweep takes the `# ci: a5`
entries, which are a5-only by their own `-p` choices.

Let a5sim take both A5 trees. The device entries drop out on the two
filters already in the loop -- they split the `__main__` sentinel and
carry `# ci: no-sim` -- which leaves the MoE, gate, expert and
transport files, each of which declares a5sim itself.

Those cannot run on torch 2.6: every MX scale is a float8_e8m0fnu
tensor, and 20 of the 24 a5sim failures in the 2026-09-19 nightly were
that one AttributeError, raised before anything compiled. Give the
a5sim leg torch 2.10, which the A5 device jobs already take; a2a3sim
keeps the default, the version the serving torch_npu pin requires.

The added cases have not been run yet -- the next nightly is their
first -- so the a5sim leg may stay red for reasons of its own. It has
been red every night regardless, at 9/33.
Stacked on #1309 (its branch is the base here); retarget to main once
that lands.


lwDavid:
Folded into #1309, which now carries both legs as one commit.

PR #1312, opened 2026-09-20
Add: wire DeepSeek V4.1 Flash prefill C1A through mHC

meromber2-eng:
- Route all C1A prefill modes through the delayed mHC pre, mix, attention, and post sequence.
- Expose hidden states, attention outputs, and next pre-mix buffers between mHC and the C1A attention leaves.
- Add distributed golden fixtures and precision checks for mHC intermediate stages, TP replication, and end-to-end output accuracy.


models/deepseek_v4_1_flash/prefill_c1a_full.py
```diff
@@ -6,374 +6,166 @@
...
-prefill_c1a_full = make_prefill_c1a_full(paged_indexer)
-prefill_c1a_full_watch = make_prefill_c1a_full(paged_indexer_direct)
+def golden_prefill_c1a_full_hc(x_hc, pre_mix, hc_attn_fn, hc_attn_scale, hc_attn_base, *attention_args):
+    """Reference the official delayed pre-mix ordering around full C1A."""
+    next_pre_mix, post_mix, residual_mix = golden_mhc_mixes(
+        x_hc, hc_attn_fn, hc_attn_scale, hc_attn_base
+    )
+    hidden = golden_mhc_pre(x_hc, pre_mix)
+    result = golden_prefill_attn_c1a_full(hidden, *attention_args)
+    output = golden_mhc_post(result.output.to(hidden.dtype), x_hc, post_mix, residual_mix)
```
high-cloud: [P2] Reduce TP attention partials before applying mHC post

The attention golden called above returns one rank's FP32 partial, but the device path performs `prefill_tp_output_all_reduce` before `mhc_post`. Consequently, this exported HC golden helper produces an incorrect full-output reference for TP2/TP4: it rounds and mixes only the local contribution. The new `golden_prefill_c1a_reindex_hc` and `golden_prefill_c1a_reuse_hc` helpers have the same issue.

A CPU-only isolated check of these helper bodies, with attention partials mocked as 1 and 2 and unit post-mix/zero residual, returns 1 for each helper, whereas the TP2 reference is 3. The CLI golden correctly uses `apply_distributed_golden`, so the passing CLI CI does not cover these helpers. Please aggregate all rank partials in FP32 before converting to BF16 and applying post, or explicitly restrict these helpers to TP1.

models/deepseek_v4_1_flash/prefill_c1a_full.py
```diff
@@ -6,368 +6,175 @@
...
+    attention_epoch: pl.Scalar[pl.INT32],
+):
+    """Collapse with the prior pre-mix, run full C1A, and produce the next pre-mix."""
+    tokens = pl.tensor.dim(x_hc, 0)
+    post_mix = pl.create_tensor([tokens, HC_MULT], dtype=pl.FP32)
+    residual_mix = pl.create_tensor([tokens, HC_MULT, HC_MULT], dtype=pl.FP32)
+    mhc_mixes(x_hc, hc_attn_fn, hc_attn_scale, hc_attn_base, next_pre_mix, post_mix, residual_mix)
+    mhc_pre(x_hc, pre_mix, hidden)
+    prefill_attn_c1a_full(
+        hidden, wq_a, wq_a_scale, q_norm_weight, wq_b, wq_b_scale, wkv, wkv_scale,
```
high-cloud: [P1] Apply attention RMSNorm between mHC collapse and the attention leaf.

The new wrapper passes the collapsed hidden directly to the attention leaf and has no attn_norm_weight parameter. The complete-layer reference in decode_layer.py (golden_decode_layer) normalizes the collapsed attention input with that weight, and the existing prefill_c2a_full.py attention_hc_pre helper implements the same order. The leaf does not perform this input normalization; its Q/KV latent norms are separate operations. Consequently, this composition cannot reproduce the model with learned attention norm weights and can change projections, index selection, and attention output. The new golden follows the same omitted-normalization path, so agreement with it does not detect the error.

Please thread attn_norm_weight through all three C1A wrappers and their fixtures, normalize after mhc_pre, and update the golden and intermediate comparisons accordingly.

models/deepseek_v4_1_flash/prefill_c1a_reuse.py
```diff
@@ -164,63 +86,61 @@ def prefill_c1a_reuse(
...
+    mhc_pre(x_hc, pre_mix, hidden)
+    prefill_attn_c1a_reuse(
+        hidden, wq_a, wq_a_scale, q_norm_weight, wq_b, wq_b_scale, wkv, wkv_scale,
+        kv_norm_weight, attn_sink, wo_a, wo_b, wo_b_scale, rope_cos, rope_sin,
+        window_slots, window_indices, window_cache, window_cache_scale,
+        compressed_cache, compressed_cache_scale, compressed_indices,
+        output_window, output_arrived, attn_out,
+        group_base, tp_rank, num_tokens, attention_epoch,
     )
+    mhc_post(attn_out, x_hc, post_mix, residual_mix, output)
```
high-cloud, line 120: [P2] Define the attention tail before mHC post consumes the full capacity.

When num_tokens is smaller than the tensor extent T, prefill_tp_output_all_reduce writes only the first num_tokens rows of attn_out, whereas mhc_post iterates over the full tensor extent. The new test/host entry declares attn_out as pure pl.Out, so the remaining rows are not initialized by the host. This makes the final HC output tail depend on allocator contents and allows NaNs to propagate. The same wiring exists in full and reindex. Current fixtures always set num_tokens equal to the tensor extent, so they do not exercise this case.

Please initialize the attention tail or restrict post-processing to the active prefix with an explicit output-tail contract, and add a case with num_tokens < T. A CPU example with a finite active prefix and a NaN attention tail confirms that mHC post propagates those NaNs; this is not an on-device reproduction.

high-cloud (commented):
Two correctness concerns in the new mHC orchestration are detailed inline. Both apply to full, reindex, and reuse. Validation was limited to static call-chain inspection and small CPU numerical examples; I did not run the NPU kernels. The A5 check was still pending at review time.

PR #1314, opened 2026-09-20
Refactor: consolidate the DSpark decode module tree

zhangqi-chen:
- Fold decode_fwd_device_state.py into decode_fwd.py, leaving a plain
  43-layer forward with no persistent-state coupling, and move its
  prepare/accept stages into decode_prepare.py
- Merge dspark_device_state.py, dspark_decode_bridge.py and
  decode_metadata.py into decode_prepare.py, which is now kernels only: the
  group prepare, metadata, RoPE gather, accept, commit, drafter-bridge and
  head-hidden fence stages
- Fold dspark_draft_step_device_state.py and dspark_proj.py into
  dspark_drafter.py, and replace dspark_prefill.py with its
  `--mode {decode,prefill}` flag
- Compose the whole recurrent step in decode_fwd_dspark.py and give it a
  Golden Harness entry point: build_tensor_specs derives 149 specs from the
  component builders, main() runs the fused L2 with no golden, and the CI
  markers request 2 devices and skip the simulator
- Build the drafter bridge tensors at the L2 frame and call
  prepare_drafter_after_target bare: rebinding its returned outputs
  versioned them inside the target scope, where the Markov sampler and the
  state commit could no longer name them
- Keep decode_fwd's tail conditional at the function frame: a pl.scope()
  around it confined the phi of the conditionally written pl.Out parameters
  to that block, out of reach of the accept stage the fused L2 runs after
  the inlined forward
- Take the per-rank token slab from MOE_TOKENS in decode_prepare and
  decode_fwd_dspark, and the drafter's query batch from the same MoE slab in
  dspark_attention: MOE_TOKENS keeps its EP=16 value, so the TP split of the
  group batch matches it only at TP 4
- Support --tp 2 / --ep 2 alongside the TP 4 deployment point in decode_fwd,
  dspark_drafter and decode_fwd_dspark, and drop the golden from decode_fwd


zhangqi-chen:
Verified on a2a3 after the fix:

| entry | tp2/ep2 | tp4/ep4 |
| --- | --- | --- |
| `decode_fwd_dspark.py` | PASS (152s) | PASS (183s) |
| `dspark_drafter.py` | PASS (47s, golden) | PASS (52s, golden) |
| `decode_fwd.py` | PASS (97s) | — |
| `dspark_markov.py --distributed` | PASS (10s, golden) | — |

On the two AI review comments: the `decode_fwd` tree in the docs no longer
lists `accept_target_into_device_state`. The fallback block-table IDs stay
shared across requests on purpose — the per-request caches in this fixture are
sized for one request's pages, so partitioning them walks off the end of the
pool, and with no golden the overlap costs nothing.

PR #1319, opened 2026-09-20
Add: keep routed expert weights packed in FP4 

### LINKED ISSUE #1287: [Feature] [API/ABI] Keep routed-expert weights packed FP4 and convert to FP8 in-kernel
### Summary

Add a supported A5/950 device ABI and kernel path that keeps DeepSeek V4.1 routed-expert weights in packed MXFP4/E2M1 form during host-to-device transfer, loads them tile by tile in the kernel, converts each tile to FP8, and reuses the existing MXFP8 matmul path.

This is a focused implementation follow-up to #1205. It covers the routed-expert weight API/ABI and kernel data path, rather than reopening the entire end-to-end MoE bring-up.

### Motivation / Use Case

The current `models/deepseek_v4_1_flash` path declares routed `w1/w2/w3` as `pl.FP8E4M3FN`. `prepare_routed_weight_for_device()` decodes the packed FP4 checkpoint weights, transposes and reorders the scales, and materializes FP8 weights on the host before the tensors are transferred to the device. `expert_routed.py` therefore invokes `pl.matmul_mx` with an already-expanded FP8 RHS.

This loses the packed checkpoint representation, creates unnecessary host work and an FP8 weight copy in HBM, and weakens the main memory-bandwidth benefit of FP4 routed experts.

A packed tensor such as `torch.float4_e2m1fn_x2` stores two FP4 values per byte. For example, physical shape `[128, 32]` represents logical shape `[128, 64]`. The logical/physical shape mapping, strides, alignment, and group-32 E8M0 scale layout must remain explicit and correct.

### Proposed API / Behavior

- Define and document the stable packed-FP4 device ABI:
  - packed E2M1 payload (`pl.FP4`);
  - physical versus logical shape;
  - even logical K/alignment requirements;
  - group-32 MX scale semantics and routed-expert transpose/layout.
- Support the minimum explicit in-kernel sequence:

  ```python
  b_fp4 = pl.load(b_packed_fp4, ...)
  b_fp8 = pl.cast(b_fp4, pl.FP8E4M3FN)
  out = pl.matmul_mx(a_fp8, a_scale, b_fp8, b_scale)
  ```

- A direct `pl.matmul_mx` overload accepting a packed-FP4 RHS is welcome, but native FP8 x FP4 MAC is not required if the explicit tile-local load/unpack/cast path is supported.
- Keep the existing FP8 input path unchanged and backward compatible.
- Do not decode or materialize the complete expert matrix on the host. Conversion should happen per loaded tile inside the kernel.
- Keep the scope on Ascend 950/A5, where the FP4 tile-ISA path is available. A2A3/910B/C fallback behavior is a separate issue and is not a blocker for this request.

### Acceptance Criteria

1. Packed `torch.float4_e2m1fn_x2` / `pl.FP4` routed-expert tiles can be loaded in JIT/kernel code, with tests for logical-to-physical shape mapping such as `[128, 32] -> [128, 64]`, strides, alignment, and invalid odd shapes.
2. Device-side FP4 unpack and FP4 E2M1 to FP8 E4M3FN conversion work for routed `w1`, `w2`, and `w3` tiles with the documented group-32 scale metadata.
3. DeepSeek V4.1 routed expert weights remain packed FP4 through host-to-device transfer; the current full host-side FP8 materialization path is no longer required for this mode and no full FP8 mirror is allocated in HBM.
4. Numerical results match an explicit unpack/dequantize/cast reference within documented tolerance, including transposed expert weights and scale ordering.
5. Add regression coverage for both the packed-FP4 path and the existing FP8 path; existing FP8 callers remain backward compatible.
6. Report the resulting HBM footprint and bandwidth/latency comparison, demonstrating removal of host-side FP8 expansion.

### Alternatives Considered

- Keeping the current host-side FP4-to-FP8 conversion is functionally correct but retains the extra FP8 materialization and HBM cost.
- Requiring native FP8 x FP4 matrix multiplication is unnecessary for the first implementation; tile-local FP4-to-FP8 conversion followed by the existing MXFP8 matmul is sufficient.
- A2A3 vector-dequant fallback should be tracked separately from the A5/950 implementation.

### Additional Context

- DeepSeek V4.1 fast-decode MoE baseline: `pypto-lib` commit `ff6db81`.
- Current implementation files: `models/deepseek_v4_1_flash/moe.py`, `models/deepseek_v4_1_flash/quantization.py`, and `models/deepseek_v4_1_flash/expert_routed.py`.
- Related: #1205 (V4.1 end-to-end MoE umbrella requirement).
- Related: #723 (previous MXFP4 expert-adoption issue, closed as stale).
- Related: https://github.com/hw-native-sys/pypto/issues/1971 (framework-side FP4 dequantization work).
- Related: https://github.com/hw-native-sys/pypto/issues/2754 (packed-FP4 tensor stride handling).

Bretty082a1:
- Keep routed W1/W2/W3 in packed FP4 tile storage with group-32 E8M0 scales and a device-side FP4 pair lookup table.
- Decode packed FP4 pairs to FP8 E4M3FN tiles inside the routed expert kernel before the existing MXFP8 matmul path.
- Add V4.1 Flash packed-weight generation, reference unpacking, and golden ABI wiring for the transposed W1/W2/W3 layouts.
- Increase the A5 ring heap to support the combined mHC and packed routed graph.

Fixes #1287


Bretty082a1:
- On a real A5 NPU with `-p a5`, EP2/TP1, `--num-tokens 8`, and the `pypto-torch214` environment, one baseline run changed from 19.33 s to 15.60 s of runtime (19.3% faster); end-to-end task time changed from 22.90 s to 19.32 s (15.6% faster).
- The routed-weight ABI footprint changed from 14.014 GB to 7.219 GB (48.5% less), including routed weights, group-32 scales, and the FP4 lookup table; this is a tensor-footprint calculation rather than total-process peak HBM.

PR #1320, opened 2026-09-21
CI: leave only V4.1-Flash on the A5 daily legs, in one table

lwDavid:
- Restrict the nightly a5sim operator sweep to DeepSeek V4.1-Flash
  and remove the DEEPSEEK_V4_VARIANT environment override.
- Combine A5 device and simulator results in one table keyed by entry,
  with pass counts and arguments for failed or unexecuted cases.
- Record every planned device case as missing before execution so an
  interrupted sweep reports incomplete coverage against the full matrix.

This removes the remaining daily V4-Pro operator sweep. The existing
DeepSeek-V4-Flash end-to-end job remains unchanged.


Hzfengsy:
@codex review

high-cloud:
Found one P2 issue in the new summary aggregation:

**A partial parameter sweep can appear as a fully passing entry.** In `v41_table.cell()` ([lines 938–941](https://github.com/hw-native-sys/pypto-lib/blob/bddfbf1c63d786d1a0e164acc16c992f6c4602a5/.github/workflows/daily_ci.yml#L938-L941)), `len(by_args) == 1` means only one result was received, not necessarily that the entry declares only one case. The workflow explicitly permits uploading partial results after an interrupted sweep.

For example, `decode_swa.py` declares six TP/DP combinations. I executed the embedded summary code with an artifact containing only:

```tsv
models/deepseek_v4_1_flash/decode_swa.py --tp 1 --dp 1	pass
```

The new table renders the entry with a bare ✅ and reports `a5 1/1 cases`, dropping the only completed configuration's arguments. The previous table retained the complete case label, so readers could see exactly which configuration passed.

Please distinguish completed results from the expected matrix and mark unexecuted configurations as missing. At minimum, retain arguments for a single parameterized result so it cannot be mistaken for a complete entry-level pass.

Otherwise, discovery matches the intended scope: 7 V4.1-Flash simulator entries remain and 33 V4-Pro entries are removed. Synthetic checks for ordinary aggregation and missing artifacts passed. No NPU tests were run for this review.

PR #1321, opened 2026-09-21
Add: GLM-5.3-Flash stream B MLA goldens and kernels

oahuohzz:
Land the Torch references and kernel bodies for all six stream B work items, each file with a script entry that proves its goldens on CPU and then validates its kernel through golden.run.

- mla_cache: per-row pl.spmd scatter into the 512-latent pool. A -1 slot owns no cache row and is skipped; the golden now skips it too instead of wrapping into the tail of the pool.
- mla_prolog: NoPE prolog as three b_trans matmuls and two rms-norm scopes; query is projected from the published, rounded q_resid. absorb_query folds the kv_b_proj key half into the query per token; absorb_output stays a weight-time golden for the loader.
- mla_epilog: row-parallel o_proj for both paths, kept FP32 as the partial sum the TP16 all-reduce consumes.
- decode_sparse_attn: 24 lanes over (token, sparse block), L1 gather of the selected latent rows, flash partials and an alpha/beta merge. A block whose lane 0 is -1 is skipped, which assumes a front-packed index list.
- prefill_sparse_attn: attends in latent space, one item per token with a pl.yield_-carried online softmax, then expands through w_v once per token. This trades two BF16 roundings against the expanded reference for keeping both expansions out of the block loop.

Both attention kernels pad the four TP16 heads to 16 zero-filled rows, so the QK/PV matmuls reach the cube minimum for L0 splitting and the FP32 reductions stay 32-byte aligned, and they reject TP < 4, where a 16-row head tile would drop heads. Script entries put the repository root first on sys.path so the golden harness is not shadowed by this directory's golden.py.

The model page records each file's state, the front-packing question for stream C, and the platform rules these kernels had to follow.


lwDavid:
建议明确保证 `topk_indices` 前置填充，或调整整块跳过逻辑：当前仅检查每个 128 项块首项，若块内后续仍有有效索引会漏算。现有测试只覆盖前置填充数据。

lwDavid (approved):

PR #1326, opened 2026-09-21
Perf: drop the redundant split_pre_post -> hc_post dependency

wangqin1723-max:
- Drop the OverlapMap dependency from split_pre_post to hc_post. post_t and
  comb_t are written by the former and read by the latter, whose ordering is
  already implied by the whole attention chain (split_pre_post ->
  mix_x_rms_norm -> ... -> proj_b_act -> hc_post); deps_viewer --edge-mode
  omitted lists exactly this pair as redundant in every HCA capture.
- Opt the two tensors out with the tensor-lifetime granularity
  (manual_dep=True). The per-argument pl.no_dep() form does not compile at
  this call site on this pin (the frontend cannot infer the callee's tensor
  metadata through the wrapper). hc_post is their only consumer, so the
  lifetime opt-out drops the same edge; a dep-gen pass confirms the pair is
  gone and the graph goes 169 -> 167 edges.

Measured on a2a3, one even die, unprofiled, 100 rounds x 3 interleaved pairs
against the pristine tree: 703.4 -> 709.0 us fast-cluster median, +5.6 us
(+0.8%), slower in 3/3 pairs, min 671.9 -> 685.1 us. Not a wall-time win: the
change is recorded here so that the graph-side lever is closed with its
measurement rather than re-tried; the other two redundant edges this graph
reports were already known not to pay.


wangqin1723-max:
Not merged — the measurement in the body is the deliverable: +5.6 us (+0.8%), slower in 3/3 interleaved pairs. Closing so the graph-side lever stays closed with its evidence rather than being re-tried; the branch remains as the record.

PR #1333, opened 2026-09-21
Perf: coarsen DSpark non-attention tiles to cut weight re-reads

wangqin1723-max:
- gate: GATE_M_TILE / GATE_N_TILE 16 -> 32 and GATE_D_TILE 2048 -> 1024,
  so the lane grid drops from 128 to 32 and the gate weight is re-read 4x
  instead of 8x; the K tile shrinks to keep the Mat tile inside its
  512 KB budget
- dspark_drafter: MAIN_PROJ_T_TILE 16 -> 128, collapsing 8 M-tiles into 1
  so the main_proj weight (K 12288 x N 4096) streams once per call
  instead of eight times
- expert_shared: allow_early_resolve on sh_gate_up_act_q and sh_w2_mm,
  the two critical-path spmd sites that lacked it

On dsv4-flash-dspark-decode-moe-ep16 (a2a3, 16 cards, 100 rounds / 5
warmup, frozen golden replayed, 3 interleaved pairs), the fastest rank's
median drops by 374.3 us (-28.3%), same sign in 3/3 pairs; golden
passes in every run. The coarser tiles change reduction order, so the
result is not bit-exact with the previous tiling.


wangqin1723-max:
Level-4 chip swimlane, before -> after, one card each (instrumented capture; the bench medians in the description are the clean numbers, so read the structure and the shares rather than the absolute makespan).

**gate** (bench 143.8 -> 69 us)

| | makespan | observed path | compute / stall |
|---|---|---|---|
| before | 171 us | ffn_norm 12.8 -> **gate_aic 128.0 (75%)** -> route_hash 6.7 | 86.5% / 13.5% |
| after | 81 us | ffn_norm 12.6 -> **gate 27.4 (34%)** -> x_norm_quant 6.8 -> gate_pre_route 1.0 (+15.1 wait) -> route_hash 7.5 | 60% / 40% |

The 128-lane grid put 75% of the stage in one 128 us AIC task; coarsening to 32 lanes takes it to 27 us.

**dspark_proj** (bench 360.5 -> 185.9 us)

| | makespan | observed path | compute / stall |
|---|---|---|---|
| before | 357 us | **dspark_main_proj_aic 342.1 (95.9%)** -> rms_norm 12.1 | 99.3% / 0.7% |
| after | 184 us | **dspark_main_proj_aic 166.2 (90.2%)** -> rms_norm 12.1 (+6.0 wait) | 96.7% / 3.3% |

Same task, half the time: 8 M-tiles re-reading the 100 MB weight became 1.

**expert_shared** (bench 81.9 -> 80.3 us)

| | makespan | first stage on path | compute / stall |
|---|---|---|---|
| before | 89 us | sh_up_mm 16.5 (+10.9 front-gap) | 55.4% / **44.6%** |
| after | 85 us | sh_gate_mm 26.9 | 74.9% / **25.1%** |

The two flags let the scheduler absorb the waits (44.6% -> 25.1% stall), but the make-span only moves 4 us, so what is left is dependency serialization across the four FFN stages.

wangqin1723-max:
Deployment-point (EP16) wall check for this PR, both arms in one 16-card allocation through the CI case path (`dsv4-flash-dspark-decode-moe-ep16`, 100 rounds / 500 warmup). The per-round headline is the max across ranks and carries the rank skew, so both arms are quoted with the same two conventions:

| arm | CI metric (fastest rank's mean) | fastest rank's median |
|---|---|---|
| before | 1529.3 us | 1441.1 us |
| after | **1188.4 us** | **969.0 us** |
| | **-22.3%** | -32.8% |

The standalone baseline measured earlier in the same configuration landed at 1529-1544 us, so the "before" arm here is consistent with it.

What the level-4 swimlane of the after arm shows: the rank spread is still large (fastest rank ~1.0 ms at 77% compute, slowest ~5-9 ms at ~86-89% core-wait), and on the slow ranks the path is dominated by `dispatch_meta_finalize` (3.7 us of work, ~5 ms of wait). That skew is a separate, larger problem than this PR addresses; the tile changes here are what move the fastest-rank number.

PR #1338, opened 2026-09-21
Add: keep routed expert weights packed in FP4

Bretty082a1:
- Store routed W1/W2/W3 as packed FP4 tiles with group-32 E8M0
  scales and a device lookup table.
- Decode packed FP4 pairs to FP8 E4M3FN tiles in the routed expert
  kernel before the existing MXFP8 matmul path.
- Add packed weight generation, reference unpacking, and V4.1 Flash
  golden ABI wiring.
- Limit FP8 staging to each active expert and skip empty experts.
- Align routed BF16 rounding and SwiGLU MXFP8 scales with CANN.


Bretty082a1:
- On a real A5 NPU with `-p a5`, EP2/TP1, `--num-tokens 8`, and the `pypto-torch214` environment, one baseline run changed from 19.33 s to 15.60 s of runtime (19.3% faster); end-to-end task time changed from 22.90 s to 19.32 s (15.6% faster).
- The routed-weight ABI footprint changed from 14.014 GB to 7.219 GB (48.5% less), including routed weights, group-32 scales, and the FP4 lookup table; this is a tensor-footprint calculation rather than total-process peak HBM.

models/deepseek_v4_1_flash/expert_routed.py
```diff
@@ -94,10 +116,122 @@ def expert_routed(
         )
         for local_i in pl.parallel(N_LOCAL_EXPERTS):
             flat_base = local_i * RECV_MAX
+            routed_w1_packed_e = routed_w1_packed[local_i, :, :]
+            routed_w3_packed_e = routed_w3_packed[local_i, :, :]
 
             n_rows = pl.read(recv_expert_count, [local_i, 0])
             n_tiles = (n_rows + RECV_TILE - 1) // RECV_TILE
 
+            w1_mxfp8 = pl.create_tensor([D, MOE_INTER], dtype=pl.FP8E4M3FN)
```
high-cloud: [P1] Bound the per-expert FP8 staging lifetime to fit the standalone heap

This inline function uses auto_scope=False, and the local-expert loop has no per-expert scope. Consequently, each W1/W3 allocation remains owned by the outer scope until it closes, even after its consumers finish. At the default EP8 configuration, W1/W3 staging alone accumulates 48 * 2 * 5120 * 2304 = 1080 MiB, exceeding the standalone runner's unchanged 1024 MiB ring_heap at line 939 before activation buffers are counted. Increasing only moe.py's heap does not cover expert_routed.py's standalone entry point. Please shorten the staging lifetime with an appropriate scope or otherwise bound the allocations, and validate the standalone entry. This is a static capacity/lifetime finding, not a reproduced NPU failure.

models/deepseek_v4_1_flash/expert_routed.py
```diff
@@ -94,10 +116,122 @@ def expert_routed(
...
+            with pl.spmd(
+                2 * (MOE_INTER // MX_MM_TASK_TILE),
+                name_hint="exp_gate_mxfp4_aiv",
+            ) as gate_mxfp4_aiv_tid:
```
high-cloud: [P2] Skip weight decoding for experts with zero received tokens

The W1/W3 decode tasks are outside the n_tiles loop, so n_rows == 0 still allocates and fully decodes both matrices. The W2 path at lines 400-407 has the same behavior. Under default EP8 this writes 48 * 3 * 5120 * 2304 bytes (about 1.58 GiB) of FP8 staging on every invocation, even with no active tokens; sparse decode therefore pays unnecessary weight reads, writes, and task scheduling for unused experts. Please guard the corresponding allocations, decode tasks, and consumers with a positive expert-count check in both phases.

high-cloud (changes requested):
Two issues remain in the new FP8 staging path: buffer lifetime exceeds the standalone runner heap budget, and empty experts still decode all weights. CPU checks passed for FP4 packing/LUT decoding, SwiGLU scale exponents, and split scale packing. No NPU run was performed for this review; the heap finding is based on allocation sizes and the scope lifetime rules documented in docs/debug-and-tune/ring-heap-and-scope-stats.md. The earlier moe/moe_test ABI mismatch is fixed in this revision.

high-cloud (approved):

PR #1344, opened 2026-09-22
Add: sequence-parallel TP wiring for the V4.1 decode attention

superwjckawaii:
- Add decode TP input AllGather and FP32-summed BF16 ReduceScatter
  over fixed token slabs, including deterministic padding and empty owners.
- Add sequence-parallel entries for C1A Full/Reindex/Reuse, SWA, C2A
  Full and C2A Reuse while retaining the replicated all-reduce entries.
- Extend decode specs, goldens and comparators for gathered inputs,
  rank-local outputs, cache state, Top-K results and combined verdicts.
- Add CPU slab and metadata helpers plus contract coverage for physical
  slab mapping and consecutive decode-layer boundaries.
- Document the compact-batch caller contract and remaining active-prefix
  and EP8 integration limits.

This implements the decode Attention sequence boundary tracked by #1275.


high-cloud (commented):
Decode Attention half-layer review (non-blocking): the fixed-width TP slab mapping is consistent for a compact batch, including T < TP and empty owners. One ABI/coverage question remains for a batch with reserved capacity greater than its active prefix. For example, with TP4, capacity=8 and num_tokens=5, each rank has two physical rows but only [2, 2, 1, 0] valid rows. `decode_common.validate_args` rejects `active_tokens < tokens` for SWA and C2A Full (C2A Reuse opts in), so those sharded half-layers are not validated for this case. The three C1A sharded test hosts instead pass `TOKENS` as `num_tokens` (e.g. `decode_c1a_full.py:646`, also Reindex/Reuse), and their fixture/golden has no separate active count; that host would process padded rows as real queries. This does not establish that the rank-level C1A kernel is wrong when called with the correct num_tokens; it is a host/golden and coverage gap. Is the intended decode caller contract that tensors are compacted every step (`tokens == active_tokens`)? If so, please document that precondition. If fixed-capacity batches are supported, please pass the active count through C1A host/golden and add sharded A5 validation for `active_tokens < tokens` across the modes, including the empty-owner case.

PR #1347, opened 2026-09-22
CI: install prometheus client for serving tests

MaxwellF1:
## Summary
- Install `prometheus-client` in the uv-backed serving CI environment.
- Install `prometheus-client` in the pip fallback serving CI environment.


MaxwellF1:
fixed in pr#1341

PR #1348, opened 2026-09-22
Fix: route local tokens through V4.1 Flash MoE

Bretty082a1:
- Remove Attention-TP owner metadata from the V4.1 Flash MoE EP ABI
- Carry per-rank token counts through dispatch, combine, and golden paths
- Handle ragged and empty local shards without treating padding as tokens
- Document the local-token contract and complete fixed-capacity validation

Related to #1275


models/deepseek_v4_1_flash/moe.py
```diff
@@ -916,72 +912,46 @@ def golden_moe(tensors):
...
+            try:
+                runtime_counts = _normalise_num_tokens(inputs["num_tokens"])
+            except (TypeError, ValueError) as exc:
+                return False, f"    invalid num_tokens: {exc}"
+            if not torch.equal(runtime_counts, counts):
+                return False, (
+                    "    num_tokens changed between fixture and comparator: "
+                    f"expected={counts.tolist()} actual={runtime_counts.tolist()}"
+                )
+        return compare_full(actual, expected, **kwargs)
```
high-cloud: [P1] Validate active token rows separately from padded capacity. `ratio_reldiff` computes its 2% bad-point budget over the entire `[EP_SIZE, MOE_TOKENS, HC_MULT, D]` tensor. With EP8 and 16 slots per rank, a run with only one active token can have that entire token row wrong while contributing just 1/128 = 0.78% of the elements; even a 50% error in every element passes `max_diff_hd=1.0`. I reproduced this with `ratio_reldiff(diff_thd=3e-3, pct_thd=0.02, max_diff_hd=1.0)`. Please compare each rank's `:counts[rank]` active prefix against its own error budget, and check the padded rows separately, so a sparse-shard routing or combine regression cannot pass validation.

PR #1352, opened 2026-09-22
Add: host-gathered FP8 partial path for the V4.1 engram kernel

supurs:
## Summary

Replaces the device-table `hash_ids` engram kernels with a single `engram` kernel that consumes host-gathered partials. The released tables (~98 GB per layer) can never sit on device, so each rank's zero-padded TP partial is gathered on the host (see the companion pypto-serving PR #259) and shipped as:

- `fp8_partial [T_DYN, 6144]` -- raw FP8E4M3FN payload, the rank's zero-padded shard flattened over hash columns x head_dim
- `scale_partial [192, T_DYN]` -- BF16-resolved E8M0 scales, transposed (the narrow-column layout is rejected by the A5 load path)

Kernel stages: (1) dequantize the 32-column groups, (2) publish via `pld.tensor.put` + `notify`, (3) chunk-wise all-reduce (`pld.tile.remote_load`) fused with the `[6144 -> 25600]` projection, (4) `engram_gate` residual. Entries `engram_rank` / `engram_group` with dynamic T; CLI `-p a5 --tp N`.

The zero-padded partial sum is bit-exact to the full-table gather (each element has exactly one non-zero contributor), so the all-reduce reproduces the released lookup exactly.

## Verification

- On-device A5 precision vs the torch golden (real-shape synthetic tables): **TP1 rel_l2 = 3.76e-08, TP2 = 9.63e-06 (2 bf16 ulp), TP4 / TP8 exact 0**.
- End-to-end with the serving host lookup over the real V4.1-Flash checkpoint (384M-row tables, real wkv/q/k weights): **TP1 / TP4 / TP8 all PASS, rel_l2 = 8.94e-07, bit-identical across TP degrees**.


models/deepseek_v4_1_flash/engram.py
```diff
@@ -280,48 +195,52 @@ def engram_tp(
...
+            lookup_partial[t0 : t0 + T_TILE, g * SCALE_GROUP : (g + 1) * SCALE_GROUP] = (
+                pl.cast(dequant, target_type=pl.BF16, mode="rint")
+            )
 
-    # Publish this rank's partial lookup into its own window slice, then barrier.
-    with pl.at(
-        level=pl.Level.CORE_GROUP, name_hint="engram_tp_publish", deps=[gather_tid]
-    ) as publish_tid:
+    # ---- Stage 2: publish the partial into this rank's window slice, barrier
+    with pl.at(level=pl.Level.CORE_GROUP, name_hint="engram_publish") as publish_tid:
```
high-cloud: [P1] `t_dim` is dynamic, but `lookup_window` has only `TP_MAX_TOKENS = 256` rows. For any input above 256 tokens the put in this stage, and the later local/remote loads, address beyond the declared window. The CLI guard only protects this standalone validation entry; it does not protect callers of the exported `engram_group`/`engram` ABI. V4.1 prefill supports thousands of tokens, so please either size the window for the supported prefill capacity or chunk the publish/reduce path, and add a device case with `T > 256`.

models/deepseek_v4_1_flash/engram.py
```diff
@@ -343,7 +262,7 @@ def engram_tp(
                 )
     with pl.at(
         level=pl.Level.CORE_GROUP,
-        name_hint="engram_tp_wait",
+        name_hint="engram_wait",
```
high-cloud: [P1] The wait in this scope uses the fixed threshold `expected=1`, which is only valid for the first execution. The signal window is updated with `AtomicAdd(1)` and is not reset, so on the second execution its value is already `>= 1`; the wait can complete before peers publish the new partial, allowing the matmul to read stale or concurrently overwritten data. Please pass an invocation epoch and wait for that monotonically increasing value (as the other V4.1 communication paths do), and test two executions of the same compiled program with different inputs.

high-cloud (changes requested):
Two correctness issues need to be addressed before this path is used by serving: the communication window does not cover the supported prefill token range, and the notification protocol is not safe across repeated executions. The current A5 tests run one 8-token invocation, so they do not exercise either case.

supurs:
Both P1 items are addressed in 9572fa9 — thanks for the detailed review.

**P1-2 (unsafe repeated execution):** adopted the same epoch protocol as `prefill_tp_output_all_reduce`. `engram` / `engram_rank` / `engram_group` now take `epoch: pl.Scalar[INT32]`; every rank bumps each peer's signal slot twice per epoch (publish, then read-done), with `wait(expected=2*(epoch-1))` headroom before this round's publish and `wait(expected=2*epoch-1)` ordering reads behind every peer's publish. A trailing read-done notify + drain wait closes the round, so the next epoch cannot overwrite a slice still being read. Added `engram_group_repeat` + per-epoch specs/golden: two epochs over one persistent window with **different** hash ids/activations, so a stale read can't hide behind identical inputs.

**P1-1 (window capacity):** `TP_MAX_TOKENS` is now 8192 — the packed-prefill chunk capacity (`DSPARK_PREFILL_MAX_TOKENS` / `PREFILL_GROUP_CAP`). We deliberately did not cap it at the DP2 balanced working point (4096): the scheduler's token budget is global across partitions, so one TP group may legally fill a whole 8192-token chunk in a single dispatch; a 4096 window would need a scheduler-side cap that does not exist today. 96 MiB/rank as BF16.

**Device tests added:** `test_large_batch_precision` (T=512, past the old 256-row cap, tp 1/4) and `test_repeat_epoch_precision` (tp 4). Local compile-only passes for tp1/tp4/repeat/T=512; the CI a5 job will pick them up.

PR #1362, opened 2026-09-23
Fix: use HC means for DSpark auxiliary target hidden states

hashiqiqixian:
## Summary

Correct the auxiliary target hidden states consumed by the DSpark drafter. In both prefill and decode, export layers 40, 41, and 42 as FP32 means across the HC streams, then cast to BF16. Keep the learned HC head, final normalization, and LM head unchanged for the target model's final output.

- Preserve TP-local layer-40/41 snapshots during prefill and read layer 42 directly; remove the auxiliary learned-head projection and its packing buffers.
- Use distinct prefill snapshot-row variable names to avoid a compiler variable-type collision.
- Align the existing prefill comparison oracle with the auxiliary HC mean. No new test files are included in this PR.

### Controlled 16-device serving A/B, 2026-09-23

**The hardware-tested candidate transplants this PR's production fix onto the pre-#1359 lib baseline; it is not a direct full-model validation of this PR's current NZ-layout-based HEAD.** The latest Serving revision used below does not yet pack DSpark weights for the layout introduced by #1359. Both A and B therefore exclude #1359; all other source versions and launch settings are held fixed.

- A: lib `3ff92ebb72271d9227c7792e12af709132d7ec0c`, immediately before #1359, without this fix.
- B: lib `97d99bf42b08649b74154449e06a824c00d10972`, A plus this PR's original two fix commits. `git range-diff` and the combined stable patch ID confirmed those transplanted patches were unchanged. The subsequent PR cleanup only removes the standalone regression test file; both model files remain identical to PR revision `17996d80`.
- One sequential allocation on the same host and devices 0-15. Each variant starts a fresh service, runs one non-streaming warmup, then one formal streaming request. Actual request concurrency is **1**, not 8 or 4.

| Metric | A: without this fix | B: with this fix | Observed change |
| --- | ---: | ---: | ---: |
| Verified-draft acceptance | 112 / 1001 = 11.1888% | 135 / 847 = 15.9386% | +4.7498 percentage points |
| Mean acceptance length, including bonus | 1.7832 | 2.1157 | +0.3325 |
| Verification rounds | 143 | 121 | -22 |
| TTFT | 368.64 ms | 328.11 ms | -40.54 ms |
| TPOT | 95.4057 ms/token | 74.6009 ms/token | -21.81% |
| Decode rate | 10.4816 tokens/s | 13.4047 tokens/s | 1.279x |
| Formal request latency | 24.6971 s | 19.3513 s | -21.65% |
| Fallbacks / errors / aborts | 0 / 0 / 0 | 0 / 0 / 0 | None |

Both variants generated identical 256-token output text, also identical to their respective warmup output. Both completed successfully, the services exited cleanly, and all 16 device locks were released.

Acceptance is calculated from the **formal-after minus warmup-after** `/metrics/json` engine counters: accepted draft tokens divided by actually verified draft tokens. Accepted tokens exclude the bonus token; verification counts are recorded before final EOS/length truncation. Mean acceptance length is `1 + accepted_tokens / speculative_drafts`. These are not the legacy runner `matched / proposed` ratios, whose scope includes initial proposals and additional asynchronous work.

TTFT is HTTP request start to the first nonempty text SSE chunk. TPOT is `(formal_request_latency - TTFT) / 255`; decode rate is `1 / TPOT`. These are client-observed serving timings, not isolated NPU kernel timings. Service startup and warmup are excluded. **This is one prompt and one formal timing per variant, not evidence of a stable aggregate speedup across workloads.**

### Environment validation

The device `hello_world` golden smoke test passed before the full-model A/B. The installed runtime build commit and PTO ISA metadata were checked against the selected PyPTO pins.

<details>
<summary>Exact source versions, environment, and startup conditions</summary>

### Source versions

| Component | Revision / version |
| --- | --- |
| pypto-serving, both variants | `bcba8d28856172e71ef421cb76213607172d6dbe` |
| pypto-lib A | `3ff92ebb72271d9227c7792e12af709132d7ec0c` |
| pypto-lib B | `97d99bf42b08649b74154449e06a824c00d10972` |
| PyPTO | `b0111c299c24ed143f8414fdd789fddb7a4270dc` |
| Simpler/runtime | `6e383fc57c12a8bd4dce045289b8b4f5d55a15f1` |
| PTO ISA | `c0d7148e95ef73bd12a73165fdce4b723a3b7e72` |
| PTOAS | Official aarch64 CPython 3.10 wheel, `v0.65`; tag commit `85af360eed58068801d21c5e2c740e14f47b146b` |
| External measurement helper | [pypto-serving test commit `bfd523ed75b75fc272074d10d318d1b243d2d015`](https://github.com/hashiqiqixian/pypto-serving/blob/bfd523ed75b75fc272074d10d318d1b243d2d015/tests/manual_dspark_screenshot_repro.py) |

The two tested transplant commits are `c86ba0147fed0cf6694b15e82867956b3d7bab33` and `97d99bf42b08649b74154449e06a824c00d10972`. They correspond to the pre-squash revisions `4ddd1187` and `17996d80`. The experiment used a separate test branch; upstream main was unchanged. After the standalone test-file cleanup, this PR was squashed into `1fdef6ff` without changing the final source tree. The helper runs outside the pristine production Serving checkout.

### Environment

| Item | Setting |
| --- | --- |
| Host / platform | Linux aarch64; NPU backend; `a2a3`; devices 0-15 |
| AICPU | Device query reports 6 available cores per device; Serving uses 4 AICPU threads |
| CANN / driver | `9.0.0` / `26.0.rc1` |
| Python / GCC | `3.10.9` / `10.3.1` |
| PyTorch / NumPy | `2.6.0+cpu` / `2.2.6` |
| Transformers / Tokenizers | `4.57.1` / `0.22.2` |
| FastAPI / Uvicorn / prometheus-client | `0.141.1` / `0.52.4` / `0.26.0` |
| Checkpoint | Same existing `dsv4-flash-0731-dspark-w8a8` checkpoint for A and B |
| CPU math threads | `OMP_NUM_THREADS=OPENBLAS_NUM_THREADS=MKL_NUM_THREADS=1` |
| Build / codegen workers | 8; runtime build targets serialized; toolchain linking limited to 1 |
| Offline execution | `HF_HUB_OFFLINE=1`, `TRANSFORMERS_OFFLINE=1`, `TOKENIZERS_PARALLELISM=false` |
| Instrumentation | CLI profiling off; `SIMPLER_DFX`, `SIMPLER_ORCH_PROFILING`, `SIMPLER_SCHED_PROFILING`, `SIMPLER_TENSORMAP_PROFILING` all 0; **`SIMPLER_HOST_STRACE=1`** |

The dedicated interpreter uses the exact-source PyPTO/runtime installations above while reusing common dependencies through layered `PYTHONPATH` entries from existing environments. This is not a fully isolated dependency lock or a complete `pip freeze`. Installed module paths and runtime build identity were checked.

Host strace remains at the runtime's default of 1: building this runtime with `SIMPLER_HOST_STRACE=0` fails because the disabled branch lacks `STRACE_DEV_SPAN_AT_A`. No runtime source patch was applied. Both variants use the same instrumentation, and `CPATH` selects the generated runtime profiling header so separately compiled orchestration sees the matching DFX configuration. This run must not be described as completely instrumentation-free.

### Launch and request settings

| Item | Setting |
| --- | --- |
| Parallelism | `--dp 4 --tp 4 --ep 16` |
| Speculation | `--speculative-config '{"method":"dspark","num_speculative_tokens":7}'` |
| Capacity | `--max-model-len 1024 --max-num-seqs 8 --max-num-batched-tokens 8192` |
| KV / prefill | `--block-size 32`; chunked prefill enabled; `--long-prefill-token-threshold 128` |
| Prefix cache | `--no-enable-prefix-caching` |
| Scheduling | Async scheduling enabled |
| CLI / prefill ring heap | `--ring-heap 2147483648,2147483648,4294967296,8589934592` (2/2/4/8 GiB) |
| Specialized ring defaults | At the pinned Serving revision: fused decode and drafter use 4/4/4/4 GiB; standalone Markov receives scalar 1 GiB. No drafter/decode-specific environment override was set. |
| Compilation | Independent build directories; no compiled-kernel cache reused between A and B |
| Endpoint / prompt | Raw `/v1/completions`, repository Palace64 prompt; no chat template |
| Sampling | Greedy: `temperature=0`, `top_p=1`, `top_k=null`, `seed=null`, `stop=[]` |
| Length / termination | 64 prompt tokens; `max_tokens=256`; the pinned completions endpoint sets `ignore_eos=True`; both finish with `length` |
| Measurement | One non-streaming warmup, then one formal SSE request, sequentially; concurrency 1 |
| Time limits | Startup 1200 s; each request 300 s; per-variant internal deadline 2100 s; device reclaim 180 s; queue job limit 5400 s |

Prompt:

```text
紫禁城南北长961米，东西宽753米，四面围有高10米的城墙，城外有宽52米的护城河。紫禁城有四座城门，南面为午门，北面为神武门，东面为东华门，西面为西华门。
```

The checkpoint config SHA256 is `a95da5598dc12d3b376fe45870dd21c16b37f98550faaf0c99acf26cfd9e20db`; the weight-index SHA256 is `81eb4ebaf6425700fab80becabfe3731b08686b4961fa281c14cfbc0dd09d7a1`. These identify the metadata, not a full bytewise checksum of all model weights.

</details>

## Scope and limitations

This PR changes only the lib-side auxiliary hidden-state export and aligns its existing prefill comparison oracle. It adds no test files and does not change the learned final-output HC head, Serving verification/sampling, Markov behavior, request scheduling, acceptance accounting, or compiler/runtime code.

Serving support for #1359's NZ/group-first weight layout, full-model validation of the current PR HEAD with that layout, and broader prompt/concurrency/repeated-timing evaluation remain outside this PR's implemented and validated scope. The measured improvement does not establish that all DSpark acceptance issues are resolved.


hashiqiqixian:
Validation evidence and version scope

The following numbers were measured in this investigation, not copied
from historical issue reports. Baseline and candidate ran sequentially
within the same eight-device allocation on host 686.

| Metric | Baseline | HC mean candidate |
| --- | ---: | ---: |
| Verification rounds | 63 | 54 |
| Verified draft tokens | 441 | 378 |
| Accepted draft tokens, excluding bonus | 64 | 73 |
| Draft acceptance rate | 14.5125% | 19.3122% |
| Mean acceptance length, including bonus | 2.01587 | 2.35185 |
| Speculative fallbacks | 0 | 0 |
| Per-position accepted counts | 41,17,3,1,1,1,0 | 40,17,8,5,2,1,0 |

Acceptance is accepted draft tokens / verified draft tokens:
`64 / 441` versus `73 / 378`, a gain of 4.7997 percentage points.
With fixed K=7, mean acceptance length is `1 + 7 * acceptance_rate`;
these are related statistics, not independent confirmations.
The baseline was also run in a separate cold launch with identical
counts, output text, and token usage.

Configuration: Ascend a2a3, devices 8-15, EP8 / TP4 / DP2, K=7,
temperature=0, prefix caching disabled, the repository Palace input
with 64 prompt tokens and 128 generated tokens, and checkpoint
`dsv4-flash-0731-dspark-w8a8`. Both runs used max-model-len 1024,
block size 32, and matching (1,1,2,4) GiB prefill and drafter ring heaps.
The returned output text was identical and both requests ended at the
length limit. This is not a raw-token-ID or reference-model oracle.
Engine counters include the final verification before output truncation.

Exact measured revisions:

- Baseline lib: `1b8caa4fd4554a7dde3317e052e52715ac51accc`.
- Candidate lib: `f9d6c9dca79f32e03bde4e8aca651e5f41afbce0`.
- Serving: `cac1a720cda7a9139b50d946d5604c03653c9abf`; its only addition
  to production revision `3f1dbd9a89d1c3a431422813f575f783e5b27e16`
  is the acceptance-metrics test helper, not a production behavior change.
  The exact [measurement helper](https://github.com/hashiqiqixian/pypto-serving/blob/cac1a720cda7a9139b50d946d5604c03653c9abf/tests/manual_dspark_acceptance.py)
  reads the serving engine metrics after the fixed cold request.
- PyPTO: `ab8e10fc3140caddfc700fa69d697148dc78c268`.
- Runtime: `6e383fc57c12a8bd4dce045289b8b4f5d55a15f1`.
- PTOAS 0.65; PTO ISA `c0d7148e95ef73bd12a73165fdce4b723a3b7e72`;
  CANN 9.0.0.

The measured candidate passed Ruff, all 10 tests in
`tests/contract/test_dspark_target_hidden.py`, actual serving prefill
and fused-decode CPU JIT/code generation, an NPU smoke test, and the
full request above. The torch tests validate tensor semantics and do
not replace PyPTO compilation or device execution.

Before opening this PR, both commits were rebased onto main
`8fb2eb930a9f94aa1bae12f7f1687af728753c12`, producing PR head
`17996d80018eb90129c7c1e81f98f6f6c26ca895`. Both patches are unchanged
according to `git range-diff`. The rebased head has not been rerun on
hardware; in particular, this does not validate integration with the
newly merged NZ-layout/weight-streaming changes. The measurements above
belong to the explicitly listed pre-rebase revisions.

Scope and limitations:

- This PR fixes the lib-side auxiliary-hidden extraction contract.
  It does not remove the target or drafter learned final HC head.
- It does not change serving scheduling, verification, or acceptance
  accounting. The serving measurement helper is outside this lib PR.
- Multi-request page-table compaction and other precision differences
  are not changed here.
- A single-prompt improvement does not establish general acceptance,
  full-model accuracy, or a stable end-to-end speedup. Acceptance is
  still low; this PR does not claim to resolve all DSpark quality gaps.

PR #1364, opened 2026-09-23
Fix: wire packed routed weights through V4.1 prefill layer

Bretty082a1:
- Pass packed MxFP4 routed weights, scales, and the pair lookup table
  through the V4.1 prefill layer MoE path
- Compact TP-owned token rows before RMSNorm and EP routing, then scatter
  the local MoE results back into the replicated prefill layout
- Align the golden fixtures and references with per-rank token counts and
  restore A5 precision coverage for TP1/DP2 and TP2/DP2


Bretty082a1:
- Wire the V4.1 prefill host, rank, and MoE sublayer entries to the
  packed MxFP4 routed-weight ABI: UINT8 W1/W2/W3 tensors, aligned MX
  scales, and the INT16 [2, 256] pair LUT.
- Apply RMSNorm before invoking the current MoE core and remove the
  stale token_owners/group_base slots from the MoE caller ABI.
- Compact TP-interleaved token rows into each rank's local prefix before
  EP routing, then scatter owned outputs back to the replicated prefill
  layout.
- Update packed routed-weight and LUT fixtures and the golden reference
  to use per-rank local token counts and the same compact/scatter layout.
- Load moe before expert_routed so transport buffer annotations use the
  layer's token extent.
Fixes #1275

